PR #4, opened 2014-04-08
Html data attributes

xificurk:
This PR (inspired by nette/nette#1399) improves handling of data attributes in `Html` objects.
- Non-scalar values are encoded using JSON in the output.
- Camel case data attribute names are automatically converted to dash separated names in the output.
- No more confusion between `$el->{'data-x'}`, `$el->data['x']` etc., both now point to the same place.
- Added getter for data attributes `$el->getData('x')`.
- The preferred way of manipulating data attributes is via properties, e.g.:

``` php
$el->data->camelCase = array();
$el->data->camelCase[] = 'item';
```

The important thing is that this change is mostly backward compatible, but there are some corner cases that won't work any more:
- Using combination of simple data attribute and named data attributes.
- The output of `$el->data['camleCase'] = 'whatever';` has changed, so if you rely on the old behavior (which you definitely should not!), you're out of luck.
- The encoding of `TRUE` and `FALSE` **for custom data attributes** has changed - the motivation for this is consistency with e.g. jQuery, so if you retrieve data using `$("#elem").data()`, you get the expected result.

There are couple of things that probably need some polishing. I'm wondering about two things (that could lead to overall simplification of the code):
1. Is the simple `data="whatever"` attribute really useful? I would gladly throw it away, but that's BC break.
2. Should the new way of manipulating data attributes be enforced more aggressively (trigger deprecation warnings)? The old way could be completely removed in the future.

### DIFF
--- a/src/Utils/Html.php
+++ b/src/Utils/Html.php
@@ -58,7 +58,7 @@ class Html extends Nette\Object implements \ArrayAccess, \Countable, \IteratorAg
 		$el->setName($parts[0]);
 
 		if (is_array($attrs)) {
-			$el->attrs = $attrs;
+			$el->addAttributes($attrs);
 
 		} elseif ($attrs !== NULL) {
 			$el->setText($attrs);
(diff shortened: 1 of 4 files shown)


dg:
I think this can be closed. Converting non-scalar values into JSON is implemented and is useful and used, `$el->data['x']` was deprecated, the encoding of TRUE and FALSE is done by `data()` and the converting to dash separated names is IMHO not needed.

PR #14, opened 2014-05-30
Validators::is() added string with pipe character

s4muel:
added pattern string with pipe (regex alternative) character


dg:
This is too hard to fix.

PR #20, opened 2014-06-09
Strings - added method cut()

icaine:
Purpose of this method is to simplify usage when functions like `strstr()` and `strrchr()` are need.


foxycode:
I'd like this:
```
after('some.long.file.name.jpg', '.', -1);
// jpg
111

PR #29, opened 2014-09-06
Help @dg as per the deal to convert tabs to spaces.

harikt:
Hi David Grudl, 

As per our conversation and your deal I am sending you my first PR to help you with the conversion.

I will start with the rest once you have merged this.

Thank you

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,40 +1,40 @@
 {
-	"name": "nette/utils",
-	"description": "Nette Utility Classes",
-	"homepage": "http://nette.org",
-	"license": ["BSD-3-Clause", "GPL-2.0", "GPL-3.0"],
-	"authors": [
-		{
-			"name": "David Grudl",
-			"homepage": "http://davidgrudl.com"
-		},
-		{
-			"name": "Nette Community",
-			"homepage": "http://nette.org/contributors"
-		}
-	],
-	"require": {
-		"php": ">=5.3.1"
-	},
-	"require-dev": {
-		"nette/tester": "~1.0"
-	},
-	"suggest": {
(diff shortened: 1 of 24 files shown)


foxycode:
I am not sure what it is going on.

@dg Do you really want to convert Nette to spaces or this is only some test?

fprochazka:
@foxycode no, it was about demonstrating how hard is it to convert properly and how much problem it creates for maintainer and contributors

harikt:
ie a truth :) . And as it was so hard to change Laravel, Cake 3 etc moved
to PSR-2 ;) .

Its all about willingness to accept than people trying to fool others.

*Hari K T*

You can ring me : +91 9388 75 8821

http://harikt.com , https://github.com/harikt ,
http://www.linkedin.com/in/harikt , http://www.xing.com/profile/Hari_KT

Skype  : kthari85
Twitter : harikt

pmjones:
Look, if you want to convert to spaces and thus being PSR-2 compliant, that's cool. If you are happy with tabs and incidentally being non-PSR-2-compliant, that's also cool. Neither is "good" or "bad" -- one is compliant, and one is not, and that's it. The tradeoffs are yours to choose between.

dg:
@pmjones :+1:

potherca:
With the risk of being slightly of topic (and assuming the PSR-2 compliance either doesn't happen or will take a while) why don't people just use git to convert from/to tabs/spaces? (We're on github, its safe to assume we're using git, right?) 

It doesn't take much more than adding 2 config settings to a project...

Full disclosure: http://stackoverflow.com/questions/2316677/can-git-automatically-switch-between-spaces-and-tabs

Issue #15, opened 2014-05-30
Validators::is() doesnt work well with pattern containing pipe character (because of string exploding)


s4muel:
i added a PR https://github.com/nette/utils/pull/14 that covers the pattern e.g.: 'pattern:(John|Carl) Doe'

because of the "foreach (explode('|', $expected) as $item)..." in the is() validator, the validator explodes the string trying to make alternatives of possibilities, but the pipe in the $expected string represents regex alternative. don't know how to solve it generally, so i made just the update to test.

dg:
This is too hard to fix.

PR #31, opened 2014-09-23
Fix #14 + #15 - Validators::is() with pipe

Majkl578:
Possible fix for #14 & #15.


dg:
Won't fix is IMHO the best solution, but thanks for PR.

PR #36, opened 2014-11-02
Arrays: added functional API

JanTvrdik:
I love those methody in JS, would be nice to have them in PHP as well. What do you think?

### DIFF
--- a/src/Utils/Arrays.php
+++ b/src/Utils/Arrays.php
@@ -229,4 +229,40 @@ class Arrays
 		return $res;
 	}
 
+	public static function findKey(array $arr, $callback)
+	{
+		foreach ($arr as $key => $value) {
+			if (call_user_func($callback, $value, $key, $arr)) {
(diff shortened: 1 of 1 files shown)


dg:
Now it can be implemented using generators:

```php
class Iterators
{
	public static function map($iterator, callable $callback)
	{
		foreach ($iterator as $key => $value) {
			yield $key => $callback($value);
		}
	}

	public static function filter($iterator, callable $callback)
	{
		foreach ($iterator as $key => $value) {
			if ($callback($value)) {
				yield $key => $value;
			}
		}
	}
```

JanTvrdik:
Yes. But I would expect `map` to return `array` if I pass `array` as argument. Using generators when `Traversable` is passed should be OK.

enumag:
@dg Generators are good but I don't think we need functions like this in Nette. Just use [nikic/iter](https://github.com/nikic/iter).

PR #51, opened 2015-01-09
Arrays.php: Remove unneccessary is_array() call

Kubo2:
We don't need to check if argument for `$arr` parameter `is_array()`, because there's a type hint for the formal parameter above in method definition.


dg:
But $arr is changes on next line.

Issue #56, opened 2015-02-19
Image::place() Unexpected output with transparency image


romanmatyus:
Code:
```php
$tux = Image::fromFile("http://upload.wikimedia.org/wikipedia/commons/a/a1/Tux2.png");
$firefox = Image::fromFile("http://people.mozilla.org/~faaborg/files/shiretoko/firefoxIcon/firefox-64.png");
$tux->place($firefox, '100%', '100%', 50);
```
should generate:
![ok](https://cloud.githubusercontent.com/assets/1110294/6276234/a65d184c-b885-11e4-9525-99046645d389.jpg)
but generate:
![fail](https://cloud.githubusercontent.com/assets/1110294/6276232/a647910c-b885-11e4-9192-53b41fb2232f.jpg)

PR #57, opened 2015-02-19
Image::place() fixed unexpected output with transparency image #56

romanmatyus:


JanTvrdik:
Do we need the three `image_copy` calls even when `$opacity` is 100 (default and most common case)?

romanmatyus:
Good point! I fixed it.

src/Utils/Image.php
```diff
@@ -487,8 +487,18 @@ public function place(Image $image, $left = 0, $top = 0, $opacity = 100)
 			);
 
 		} elseif ($opacity <> 0) {
+			$cutting = imagecreatetruecolor($this->getWidth(), $this->getHeight());
```
JanTvrdik: This should be `imagecreatetruecolor($image->getWidth(), $image->getHeight())`

src/Utils/Image.php
```diff
@@ -487,8 +487,18 @@ public function place(Image $image, $left = 0, $top = 0, $opacity = 100)
...
+			imagecopy(
+				$cutting, $this->image,
+				0, 0, $left, $top, $this->getWidth(), $this->getHeight()
```
JanTvrdik: This should be `0, 0, $left, $top, $image->getWidth(), $image->getHeight()`

romanmatyus:
@JanTvrdik Thanks! Now it's many times faster.

dg:
Thanks!

PR #58, opened 2015-02-26
Workaround (hopefully temporary) for changed signature of DateTime::createFromFormat in PHP7

JanTvrdik:
Changed signature of DateTime::createFromFormat in PHP7 makes a lot of Nette tests fail.

We may
A) ignore it for now
B) workaround it
C) remove the method entirely

### DIFF
--- a/src/Utils/DateTime.php
+++ b/src/Utils/DateTime.php
@@ -10,6 +10,9 @@ namespace Nette\Utils;
 use Nette;
 
 
+// workaround for changed signature of createFromFormat in PHP 7
+@call_user_func(function() {
+
 /**
  * DateTime.
  *
@@ -125,3 +128,5 @@ class DateTime extends \DateTime
 	}
 
 }
+
+});


JanTvrdik:
Can this be merged?

dg:
I think there is no need to do a workaround for PHP release, which is not yet in beta.

hrach:
Well, this bug may cover other incompatibilities. No need to merge it to stable, but master should/could be ok for this.

dg:
Another workaround:

composer.json:
```
	"autoload": {
		"classmap": ["src/"],
		"files": ["src/php7-workaround.php"]
	},
```

php7-workaround.php:
```php
<?php
@class_exists('Nette\Utils\DateTime');
```

JanTvrdik:
Nice, but I like my workaround more because it is located in a single file which is also the source of the issue and does not force loading of two files which you may not need.

JanTvrdik:
I reworked your idea to a single file =)

dg:
what about `call_user_func(function(){...})` to not polute global space?

dg:
Reverted in PHP https://github.com/php/php-src/commit/b148c50f4864684aa9a6c6483cc309a7f2badbb5 (but exists in 7.0.0 alpha1).

JanTvrdik:
I know. But there is currently no [snapshot](http://windows.php.net/downloads/snaps/master/) available to confirm that it works. Hopefully they will fix the snapshots soon.

dg:
:+1:

PR #61, opened 2015-03-30
Json: add fix for Json::decode() collision attack

fabik:
If you are using Json::decode() in your application on a user input, it is prone to denial-of-service (DoS) attacks. The worst case time complexity of json_decode() is O(n^2), where n is the number of elements in the object. The test case in this pull request creates a 1,6 megabyte JSON string that takes about 1 minute to decode.

This could be fixed using randomization. If the JSON string is longer than ~50000 characters, we can prepend all string literals with ~3 random characters using a regexp. Then we can iterate over the returned arrays, objects and strings and remove the prepended characters from them. We can measure time from the beginning of the call of the Json::decode() method every ~1000 iterations. If it is longer than ~1 second, we can throw an exception.

Do you think it is a good idea?

### DIFF
--- a/src/Utils/Json.php
+++ b/src/Utils/Json.php
@@ -98,6 +98,65 @@ class Json
 		return $value;
 	}
 
+
+	/**
+	 * Safely decodes a JSON string.
+	 * @param  string
+	 * @param  int  accepts Json::FORCE_ARRAY
+	 * @param  float
+	 * @return mixed
+	 */
+	public static function decodeSafe($json, $options = 0, $timeLimit = 2.0)
+	{
+		$timeLimit += microtime(true);
+
+		$json = preg_replace_callback('#"(?:\\\\.|[^"\\\\])*"#', function($m) use (&$salt) {
+			if ($salt === null || !mt_rand(0, 100)) {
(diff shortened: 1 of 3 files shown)


fprochazka:
Implementing a custom (alternative) json parser with configurable time limit might solve that be nice. Something like "try to parse this, but if it takes more than this time throw an exception". I don't think it should be replacement of the current json::decode, but rather an addition to it, in a new method.

JanTvrdik:
This is a know issue which can not be solved in user-land without implementing custom hash map with randomized hash function.

Some ugly workarounds (such as your proposal) are possible but none of them is (AFAIK) nice enough to be included in Nette.

fabik:
I agree it's ugly, but all I want is to sleep calmly that my application cannot be shut down so easily. The class Nette\Utils\Json contains only a set of workarounds for internal json_*() functions, so I thought it would fit in there. Do you think I should create a separate project for that instead?

fabik:
Here we go. I know it's shit. But it works.

fabik:
This method has a large overhead. Therefore, I'm closing it.

PR #62, opened 2015-04-07
Fix Arrays::associate if $rowOrig is Traversable

Vyki:
Arrays::associate fails if $rowOrig is not array or instance of stdClass.


Vyki:
It solves also http://forum.nette.org/cs/22696-fetchassoc-hazi-notice-undefined-index

JanTvrdik:
See also https://github.com/nette/database/pull/61

Vyki:
I think that the [nette/database#61](https://github.com/nette/database/pull/61) is only nice workaround.

dg:
None of the functions in Arrays support Traversable, so  nette/database#61 seems to me better.

PR #63, opened 2015-05-28
DateTime @return anotation changed to "static"

Achse:
- feature
- issues
- documentation - not needed
- BC break - no

**Why this?** 
When I inherit `Nette\DateTime` and add `xyz` method to this child-class, and then I create new instace like:

```php
$dateTime = MyDateTime::from('2015-12-24 12:00:00');
$dateTime->xyz();  // <-- here, IDE assume its Nete\DateTime 
                   // where no xyz method has been declared.
```

See: http://www.phpdoc.org/docs/latest/guides/types.html

### DIFF
--- a/src/Utils/DateTime.php
+++ b/src/Utils/DateTime.php
@@ -42,7 +42,7 @@ class DateTime extends \DateTime implements \JsonSerializable
 	/**
 	 * DateTime object factory.
 	 * @param  string|int|\DateTime
-	 * @return DateTime
+	 * @return static
 	 */
 	public static function from($time)
 	{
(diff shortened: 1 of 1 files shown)


xificurk:
http://forum.nette.org/cs/21479-jak-funguje-return-this-ve-vasem-ide

dg:
You should use `self` just to be consistent with rest of framework.

Achse:
@dg Im not sure if we understand each other. `self` vs. `static` is not about consistency but it's about correctly describing behavior of the method.

- There is `return new static($time);` so it returns instance of itself (polymorphim modifies that). And there shoud be `@return static` anotation to describe this behaivour.

- If it would be `return new self($time);` or `return new DateTime($time);` instead, the `@return self` whould be correct.

JanTvrdik:
@Achse Read once again the forum thread linked above. What you describe is a theory which does not work reliably in practice.

PR #64, opened 2015-06-08
PHP7: minor fixes for tests

JanTvrdik:


JanTvrdik:
Finally green on PHP 7!

PR #66, opened 2015-06-13
Random: use random_bytes() on PHP 7

JanTvrdik:
Not sure whether the generateBytes() refactoring is worth it. It's still a bit ugly.

### DIFF
--- a/src/Utils/Random.php
+++ b/src/Utils/Random.php
@@ -26,6 +26,10 @@ class Random
 	 */
 	public static function generate($length = 10, $charlist = '0-9a-z')
 	{
+		if ($length === 0) {
+			return ''; // mcrypt_create_iv does not support zero length
+		}
+
 		$charlist = str_shuffle(preg_replace_callback('#.-.#', function($m) {
 			return implode('', range($m[0][0], $m[0][2]));
 		}, $charlist));
--- a/src/Utils/Random.php
+++ b/src/Utils/Random.php
@@ -27,7 +27,7 @@ class Random
 	public static function generate($length = 10, $charlist = '0-9a-z')
 	{
 		if ($length === 0) {
-			return ''; // mcrypt_create_iv does not support zero length
+			return ''; // random_bytes and mcrypt_create_iv do not support zero length
 		}
 
 		$charlist = str_shuffle(preg_replace_callback('#.-.#', function($m) {
@@ -35,7 +35,10 @@ class Random
 		}, $charlist));
 		$chLen = strlen($charlist);
 
-		if (function_exists('openssl_random_pseudo_bytes')) {
+		if (PHP_VERSION_ID >= 70000) {
+			$rand3 = random_bytes($length);
+		}
+		if (empty($rand3) && function_exists('openssl_random_pseudo_bytes')) {
 			$rand3 = openssl_random_pseudo_bytes($length);
 		}
 		if (empty($rand3) && function_exists('mcrypt_create_iv')) {
--- a/tests/Utils/Random.generate().no-mcrypt.phpt
+++ b/tests/Utils/Random.generate().no-mcrypt.phpt
@@ -1,7 +1,8 @@
 <?php
 
 /**
- * Test: Nette\Utils\Random::generate()
+ * Test: Nette\Utils\Random::generate() without openssl and mcrypt
+ * @phpIni disable_functions=openssl_random_pseudo_bytes,mcrypt_create_iv
  */
 
 use Nette\Utils\Random,
@@ -9,6 +10,11 @@ use Nette\Utils\Random,
 
 
 require __DIR__ . '/../bootstrap.php';
+
+
+if (!extension_loaded('mcrypt')) {
+	Tester\Environment::skip('Requires PHP extension mcrypt.');
+}
 
 
 Assert::same( 10, strlen(Random::generate()) );
--- a/tests/Utils/Random.generate().no-openssl.phpt
+++ b/tests/Utils/Random.generate().no-openssl.phpt
@@ -1,7 +1,8 @@
 <?php
 
 /**
- * Test: Nette\Utils\Random::generate()
+ * Test: Nette\Utils\Random::generate() without openssl
+ * @phpIni disable_functions=openssl_random_pseudo_bytes
  */
 
 use Nette\Utils\Random,
@@ -9,6 +10,11 @@ use Nette\Utils\Random,
 
 
 require __DIR__ . '/../bootstrap.php';
+
+
+if (!extension_loaded('openssl')) {
+	Tester\Environment::skip('Requires PHP extension OpenSSL.');
+}
 
 
 Assert::same( 10, strlen(Random::generate()) );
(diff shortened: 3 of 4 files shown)


JanTvrdik:
Also: both `random_bytes` and `mcrypt_create_iv` emit warnings if `$length` is zero.

src/Utils/Random.php
```diff
@@ -52,10 +43,41 @@ public static function generate($length = 10, $charlist = '0-9a-z')
...
+			$bytes = openssl_random_pseudo_bytes($length);
+		} elseif ($strategy === 2 && function_exists('mcrypt_create_iv')) {
+			$bytes = mcrypt_create_iv($length, MCRYPT_DEV_URANDOM);
+		} elseif ($strategy === 3 && @is_readable('/dev/urandom')) {
+			$bytes = file_get_contents('/dev/urandom', FALSE, NULL, -1, $length);
+		} elseif ($strategy === 4) {
+			static $cache;
+			$bytes = str_pad('', $length, $cache ?: $cache = md5(serialize($_SERVER), TRUE));
+		} elseif ($strategy === 5) {
+			throw new Nette\InvalidStateException();
```
fprochazka: you might wanna add here some explanatory message

src/Utils/Random.php
```diff
@@ -52,10 +43,41 @@ public static function generate($length = 10, $charlist = '0-9a-z')
...
+		} elseif ($strategy === 1 && function_exists('openssl_random_pseudo_bytes')) {
+			$bytes = openssl_random_pseudo_bytes($length);
+		} elseif ($strategy === 2 && function_exists('mcrypt_create_iv')) {
+			$bytes = mcrypt_create_iv($length, MCRYPT_DEV_URANDOM);
+		} elseif ($strategy === 3 && @is_readable('/dev/urandom')) {
+			$bytes = file_get_contents('/dev/urandom', FALSE, NULL, -1, $length);
+		} elseif ($strategy === 4) {
+			static $cache;
+			$bytes = str_pad('', $length, $cache ?: $cache = md5(serialize($_SERVER), TRUE));
+		} elseif ($strategy === 5) {
```
fprochazka: this should probably by >= 5

enumag: I think it should be `} else {`.

dg: This is really magic implementation

hrach:
I would use while cycle, no need for recursion.

src/Utils/Random.php
JanTvrdik: That's why it is private. I originally wanted to make the method public.

JanTvrdik:
I hate both the original and my own implementation :disappointed:

dg:
Whats wrong with current implementation? It's straightforwardhttps://github.com/nette/utils/blob/master/src/Utils/Random.php#L34-L45.

JanTvrdik:
> Whats wrong with current implementation? 

You evaluate `empty($rand3)` more times than it is necessary. Unfortunately every alternative implementation I have came up with either had similar problem or was too complicated.

JanTvrdik:
We currently have inconsistency between `php.ini` for Windows and Linux. Should this be fixed as well?

dg:
Of course, but are they really inconsistent?

JanTvrdik:
> but are they really inconsistent?

That's what I don't know. Are some extensions on Linux available even with `-n`? 

* `php-unix.ini` enables JSON
* `php-win.ini` enabled mbstring, gd2, fileinfo and openssl

dg:
Are tests failing?

JanTvrdik:
No, but in `Random` case if you don't have openssl extension, you will just not be able to test the `openssl_random_pseudo_bytes` strategy. You would need to generate code coverage on CI to see what has actually been covered.

dg:
Random is used to work in all environments, so it is really hard to test it completely.

Btw I don't see openssl here https://github.com/nette/utils/blob/master/tests/php-win.ini

JanTvrdik:
@dg Yes, I added openssl in the first commit.

> it is really hard to test it completely

Yes, but with the improved tests (see the 1st commit) it tests first 4 strategies on Travis. If you run them on Windows, it will test the last strategy as well.

milo:
> Are some extensions on Linux available even with -n? 

These are compile args from Debian rules files for PHP 5.6.7 cli. All others extensions must be installed as a `.so` library.
```
COMMON_CONFIG=--build=$(DEB_BUILD_GNU_TYPE) \
                --host=$(DEB_HOST_GNU_TYPE) \
                --sysconfdir=/etc \
                --localstatedir=/var \
                --mandir=/usr/share/man \
                --disable-debug \
                --with-regex=php \
                --disable-rpath \
                --disable-static \
                --with-pic \
                --with-layout=GNU \
                --with-pear=/usr/share/php \
                --enable-calendar \
                --enable-sysvsem \
                --enable-sysvshm \
                --enable-sysvmsg \
                --enable-bcmath \
                --with-bz2 \
                --enable-ctype \
                --with-db4 \
                --with-qdbm=/usr \
                --without-gdbm \
                --with-iconv \
                --enable-exif \
                --enable-ftp \
                --with-gettext \
                --enable-mbstring \
                --with-onig=/usr \
                --with-pcre-regex=/usr \
                --enable-shmop \
                --enable-sockets \
                --enable-wddx \
                --with-libxml-dir=/usr \
                --with-zlib \
                --with-kerberos=/usr \
                --with-openssl=/usr \
                --enable-soap \
                --enable-zip \
                --with-mhash=yes \
                --with-system-tzdata \
                --with-mysql-sock=/var/run/mysqld/mysqld.sock \
                $(CONFIGURE_DTRACE_ARGS)

CFLAGS="$(CFLAGS)" PROG_SENDMAIL="$(PROG_SENDMAIL)" ../configure \
                --prefix=/usr --disable-cgi \
                --with-config-file-path=/etc/php5/cli \
                --with-config-file-scan-dir=/etc/php5/cli/conf.d \
                $(COMMON_CONFIG) \
                --without-mm \
                --disable-pdo \
                --without-mysql --without-sybase-ct \
                --without-mssql --without-sqlite3 --enable-pcntl \
                --with-libedit=shared
```

JanTvrdik:
@dg I think this is ready to be merged. Let me know if there is anything that should be improved.

tests/Utils/Random.generate().no-mcrypt.phpt
```diff
@@ -0,0 +1,29 @@
...
+if (!extension_loaded('mcrypt')) {
+	Tester\Environment::skip('Requires PHP extension mcrypt.');
+}
+
+
+Assert::same( 10, strlen(Random::generate()) );
+Assert::same( 0, strlen(Random::generate(0)) );
+Assert::same( 5, strlen(Random::generate(5)) );
+Assert::same( 200, strlen(Random::generate(200)) );
+
```
dg: Can you add `Assert::random()` test?

dg:
Thanks!

tests/Utils/Random.generate().no-mcrypt.phpt
JanTvrdik: Sorry, I don't understand what you mean.

PR #70, opened 2015-07-15
Json: accept whitespace-surrounded "null" for decode()

ondrej-bouda:
Json: accept whitespace-surrounded `"null"` for `decode()` as it is a valid JSON text. As the standard says: "Insignificant whitespace is allowed before or after any token." Therefore, `" null   "` and others are legal, although `Json` throws an exception for such input.

### DIFF
--- a/src/Utils/Json.php
+++ b/src/Utils/Json.php
@@ -89,7 +89,7 @@ class Json
 		}
 		$value = call_user_func_array('json_decode', $args);
 
-		if ($value === NULL && $json !== '' && strcasecmp($json, 'null')) { // '' is not clearing json_last_error
+		if ($value === NULL && $json !== '' && strcasecmp(trim($json, " \t\n\r"), 'null') !== 0) { // '' is not clearing json_last_error
 			$error = json_last_error();
 			throw new JsonException(isset(static::$messages[$error]) ? static::$messages[$error] : 'Unknown error', $error);
 		}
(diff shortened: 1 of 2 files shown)


src/Utils/Json.php
```diff
@@ -89,7 +89,7 @@ public static function decode($json, $options = 0)
 		}
 		$value = call_user_func_array('json_decode', $args);
 
-		if ($value === NULL && $json !== '' && strcasecmp($json, 'null')) { // '' is not clearing json_last_error
+		if ($value === NULL && $json !== '' && strcasecmp(trim($json, " \t\n\r"), 'null') != 0) { // '' is not clearing json_last_error
```
dg: `!= 0` is not needed

ondrej-bouda: Indeed, but it seems more readable to me.

dg: Anyway, it must be written with `===`

ondrej-bouda: `strcasecmp()` always returns `int` (or it is supposed so, at least), so there is no difference between using `==` and `===` in this case.

Majkl578: Considering larger input, wouldn't `preg_match` be a better choice performance-wise?
```php
!preg_match('~^\s*null\s*$~is', $json)
```

http://3v4l.org/X9X9u

ondrej-bouda: I considered using `preg_match()`, but I decided using `trim()` would be better, because it is faster for the valid case (i.e., when the decoded JSON was a correct `"null"`), while it is slightly slower in the erroneous case. I believe this is the correct decision.

dg: It is coding style. You must use `===` everywhere.

dg:
please squash it to single commit

enumag:
bump

PR #72, opened 2015-08-03
Image: fix exception message [Closes #71]

### LINKED ISSUE #71: Exception string: Unsupported image type '$type'.
Hi,

I've just got weird message, while saving image. It seems like typo.
`throw new Nette\InvalidArgumentException('Unsupported image type \'$type\'.')`

So I would like to fix & pull, but suggested solution is not good enough. When user uploads image "example.jpegg" variable $type is empty and message is: Unsupported image type ''.
`throw new Nette\InvalidArgumentException("Unsupported image type '$type'.");`

Can somebody (with more experience) take this issue? 
Thanks

enumag:


src/Utils/Image.php
```diff
@@ -530,7 +530,7 @@ public function save($file = NULL, $quality = NULL, $type = NULL)
 				return imagegif($this->image, $file);
 
 			default:
-				throw new Nette\InvalidArgumentException('Unsupported image type \'$type\'.');
+				throw new Nette\InvalidArgumentException($type ? "Unsupported image type \'$type\'." : 'Unsupported file extension \'' . pathinfo($file, PATHINFO_EXTENSION) . '\'.');
```
Majkl578: What about throwing about unsupported extension in the switch above in default branch (https://github.com/nette/utils/pull/72/files#diff-73701c083ecb416c68fb31139c302d43R530)?

enumag: Good idea.

src/Utils/Image.php
```diff
@@ -530,7 +532,7 @@ public function save($file = NULL, $quality = NULL, $type = NULL)
 				return imagegif($this->image, $file);
 
 			default:
-				throw new Nette\InvalidArgumentException('Unsupported image type \'$type\'.');
+				throw new Nette\InvalidArgumentException("Unsupported image type \'$type\'.");
 		}
```
dg: `\'` is not needed, just use `'`

enumag: Thanks, fixed.

dg:
Great. Can you add test?

src/Utils/Image.php
```diff
@@ -514,6 +514,8 @@ public function save($file = NULL, $quality = NULL, $type = NULL)
 					break;
 				case 'gif':
 					$type = self::GIF;
+				default:
+					throw new Nette\InvalidArgumentException('Unsupported file extension \'' . pathinfo($file, PATHINFO_EXTENSION) . '\'.');
```
Majkl578: You can also rewrite this with double quotes so you don't have to escape, also you can reuse pathinfo output from the switch above?
```php
switch (strtolower($ext = pathinfo($file, PATHINFO_EXTENSION))) {
```

enumag: I know, I'm just not sure if it is ok for Nette. @dg ?

dg: Sure, it is ok.

src/Utils/Image.php
```diff
@@ -574,8 +576,8 @@ public function __toString()
 	 */
 	public function send($type = self::JPEG, $quality = NULL)
 	{
-		if ($type !== self::GIF && $type !== self::PNG && $type !== self::JPEG) {
-			throw new Nette\InvalidArgumentException('Unsupported image type \'$type\'.');
+		if (!in_array($type, [self::JPEG, self::PNG, self::GIF], TRUE)) {
```
enumag: Just trying to unify this with fromString method. Is it ok?

enumag:
Tests added.

dg:
Thank you!

PR #73, opened 2015-08-03
Image: add support for webp

enumag:
Not sure how to write tests - there aren't any tests specific for the other types either.

Also should I check for php version?

### DIFF
--- a/src/Utils/Image.php
+++ b/src/Utils/Image.php
@@ -101,7 +101,8 @@ class Image extends Nette\Object
 	/** image types */
 	const JPEG = IMAGETYPE_JPEG,
 		PNG = IMAGETYPE_PNG,
-		GIF = IMAGETYPE_GIF;
+		GIF = IMAGETYPE_GIF,
+		WEBP = -1;
 
 	const EMPTY_GIF = "GIF89a\x01\x00\x01\x00\x80\x00\x00\x00\x00\x00\x00\x00\x00!\xf9\x04\x01\x00\x00\x00\x00,\x00\x00\x00\x00\x01\x00\x01\x00\x00\x02\x02D\x01\x00;";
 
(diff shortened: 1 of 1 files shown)


src/Utils/Image.php
```diff
@@ -101,7 +101,8 @@ class Image extends Nette\Object
 	/** image types */
 	const JPEG = IMAGETYPE_JPEG,
 		PNG = IMAGETYPE_PNG,
-		GIF = IMAGETYPE_GIF;
+		GIF = IMAGETYPE_GIF,
+		WEBP = -1;
```
enumag: Similar constant for WEBP does not exist (yet).

src/Utils/Image.php
```diff
@@ -576,10 +583,10 @@ public function __toString()
 	 */
 	public function send($type = self::JPEG, $quality = NULL)
 	{
-		if (!in_array($type, [self::JPEG, self::PNG, self::GIF], TRUE)) {
+		if (!in_array($type, [self::JPEG, self::PNG, self::GIF, self::WEBP], TRUE)) {
 			throw new Nette\InvalidArgumentException("Unsupported image type '$type'.");
 		}
-		header('Content-Type: ' . image_type_to_mime_type($type));
+		header('Content-Type: ' . ($type === self::WEBP ? 'image/webp' : image_type_to_mime_type($type)));
```
enumag: Webp is not supported by image_type_to_mime_type.

dg:
Does getimagesizefromstring support WEBP?

enumag:
Oh boy of course it does not. :-(

enumag:
There is no way to detect webp images in fromString and fromFile (aside from checking file extension in fromFile) so I removed webp from both. But I still think Nette\Utils\Image should be able to generate webp images.

dg:
It is possible via `$image->webp('file.webp');` Full support is currently IMHO not possible.

enumag:
Ok, closing for now. Thank you.

enumag:
Webp support has been added for PHP 7.1RC4. It should be possible to implement this now.

PR #74, opened 2015-08-03
Mapper: simplify constructor

enumag:


JanTvrdik:
Reasoning? BC impact?

enumag:
`Nette\Utils\Callback::check()` just checks if the argument is callable and throws an exception otherwise which is pretty much the same as what callable type hint does and since we require 5.4 now we can simplify it. Unless someone was catching that exception (which I highly doubt) there should be no BC break.

src/Iterators/Mapper.php
```diff
@@ -19,10 +19,10 @@ class Mapper extends \IteratorIterator
 	private $callback;
 
 
-	public function __construct(\Traversable $iterator, $callback)
+	public function __construct(\Traversable $iterator, callable $callback)
```
fprochazka: http://phpfashion.com/php-5-4-ma-nepouzitelny-typ-callable

enumag: Yes I read that. Irrelevant in this case.

dg:
what about `Mapper::__construct added typehint callable`?

enumag:
@dg Done.

@fprochazka If I understand the article correctly, it basically says to use `Nette\Utils\Callback::check($callback, TRUE);` instead of `Nette\Utils\Callback::check($callback, TRUE);` or callable typehint. If @dg wants that instead I don't mind.

PR #76, opened 2015-08-14
PHP7 scalar and return type hints support

kukulich:


enumag:
Maybe the param and return annotations should be removed as they're pretty much duplicate.

kukulich:
I agree with you. I've prepared this principally to discuss how the PHP7 pull requests should be made.

Majkl578:
> how the PHP7 pull requests should look

What do you mean? PHP 7 changes are backward incompatible and there is no way to make it cross-compatible with PHP 5.x. The only way seems to be some sort of converter like we had with 5.2 vs. 5.3. This would be PITA.

> Maybe the param and return annotations should be removed as they're pretty much duplicate.

They are not always, because PHP 7 still does not support nullable types. Also Nette heavily uses mixed, sadly.

src/Utils/ArrayHash.php
```diff
@@ -21,7 +23,7 @@ class ArrayHash extends \stdClass implements \ArrayAccess, \Countable, \Iterator
 	 * @param  bool
 	 * @return self
 	 */
-	public static function from($arr, $recursive = TRUE)
+	public static function from(array $arr, bool $recursive = TRUE): self
```
Majkl578: static

src/Utils/DateTime.php
```diff
@@ -39,7 +41,7 @@ class DateTime extends \DateTime implements \JsonSerializable
 	 * @param  string|int|\DateTime
 	 * @return self
 	 */
-	public static function from($time)
+	public static function from($time): self
```
Majkl578: static

kukulich:
> What do you mean? PHP 7 changes are backward incompatible and there is no way to make it cross-compatible with PHP 5.x. The only way seems to be some sort of converter like we had with 5.2 vs. 5.3. This would be PITA.

Exactly. So there has to be some discussion if PHP 7 features will be in new branch, in master branch but in comments or there will be some sort of converter.

There should probably be special branch for PHP 7 because of composer. Version 2.4 can be PHP 5.4 compatible, version 3.0 for PHP 7+.

JanTvrdik:
:-1: Pointless, premature. Author does not give any arguments why should this PR be of any value.

kukulich:
https://twitter.com/geekovo/status/631192985179410433

dg:
The complicated part is that now you have to rebase this PR until Nette goes to PHP 7.

dg:
Btw, interesting thing is that return `self` works as `static` and `static` is parse error http://3v4l.org/TvJjC. This is the end of discussion about `return @self`.

tests/Utils/Image.send.phpt
```diff
@@ -20,5 +22,5 @@ $main = Image::fromFile('images/logo.gif');
 
 
 Assert::exception(function () use ($main) { // invalid image type
-	$main->send('jpg');
-}, 'Nette\InvalidArgumentException', 'Unsupported image type \'jpg\'.');
+	$main->send(IMG_WBMP);
```
dg: Why this change?

kukulich: The parameter should be integer, string `jpg` throws `TypeError`. However I can add another test for the `TypeError` exception.

dg: Oh, I understand

kukulich:
> The complicated part is that now you have to rebase this PR until Nette goes to PHP 7.

Yes. However I think there are plenty of projects that would use PHP 7 Nette (eg. Slevomat, Rohlik, DameJidlo) so there should be enough people to do the rebase.

kukulich:
I will prepare another pull request with changes that could be merged to master so the rebase would be easier then.

JanTvrdik:
The rebasing has to be fully or at least partially automated. Doing it manually would by boring and error-prone. @kukulich Do you have time to work on such tool?

kukulich:
There's already tool doing the opposite: https://github.com/ondrejbouda/php7backport

So if we move master to PHP 7, we can rebase another branch to be PHP 5.4 compatible.

I think we can prepare automatic tool during our hackathons in Slevomat.

JanTvrdik:
> So if we move master to PHP 7

Not going to happen any time soon. Also the tool is currently missing a lot of conversions and the conversions that are present are often not implemented properly.

dg:
To move Nette to PHP 7 means

- move it to PHP 5.5 and to 5.6 (see end of https://phpfashion.com/jak-se-opousti-php-5-3)
- have tools like ApiGen ready for PHP 7 syntax
- have the ability to test it on PHP 7 (it means do it at the time when PHP7 will be released)

kukulich:
PHP 7 is not released so it's time to improve the tool :)

I think it doesn't matter if master is in PHP 7 or not. It just depends how the release process works.

dg:
@kukulich Do you have any tool to generate typehints? And could you send PR to other packages too?

kukulich:
@dg It should be possible with this: https://github.com/slevomat/coding-standard/blob/2.0-dev/SlevomatCodingStandard/Sniffs/TypeHints/TypeHintDeclarationSniff.php

There's no documentation but I can prepare a manual for you how to use it next week.

dg:
Great!

kukulich:
@dg https://github.com/kukulich/php-type-hints-convertor

I hope it will help.

dg:
Thanks :-)

PR #77, opened 2015-08-16
Compatibility with PHP 7 strict mode

kukulich:


src/Utils/Image.php
```diff
@@ -321,15 +321,15 @@ public function resize($width, $height, $flags = self::FIT)
 	 */
 	public static function calculateSize($srcWidth, $srcHeight, $newWidth, $newHeight, $flags = self::FIT)
 	{
-		if (substr($newWidth, -1) === '%') {
-			$newWidth = round($srcWidth / 100 * abs($newWidth));
+		if (substr((string) $newWidth, -1) === '%') {
```
JanTvrdik: Maybe `is_string($newWidth) && substr(...)`?

src/Utils/Image.php
```diff
@@ -321,15 +321,15 @@ public function resize($width, $height, $flags = self::FIT)
...
+			$newWidth = (int) round($srcWidth / 100 * abs($newWidth));
 			$percents = TRUE;
 		} else {
 			$newWidth = (int) abs($newWidth);
 		}
 
-		if (substr($newHeight, -1) === '%') {
-			$newHeight = round($srcHeight / 100 * abs($newHeight));
+		if (substr((string) $newHeight, -1) === '%') {
+			$newHeight = (int) round($srcHeight / 100 * abs($newHeight));
```
JanTvrdik: Casting `round` to int? Why not just cast the expression, i.e. `(int) ($srcHeight / 100 * abs($newHeight))`?

src/Utils/Strings.php
```diff
@@ -205,7 +205,7 @@ public static function webalize($s, $charlist = NULL, $lower = TRUE)
 		if ($lower) {
 			$s = strtolower($s);
 		}
-		$s = preg_replace('#[^a-z0-9' . preg_quote($charlist, '#') . ']+#i', '-', $s);
+		$s = preg_replace('#[^a-z0-9' . preg_quote((string) $charlist, '#') . ']+#i', '-', $s);
```
JanTvrdik: Maybe avoid calling `preg_quote` for `NULL` value?

src/Utils/Image.php
kukulich: `(int) 1.9` would become `1` then so it's better to round it first.

JanTvrdik: Maybe `(int) ($srcHeight / 100 * abs($newHeight) + 0.5)`?

src/Utils/Strings.php
kukulich: Make sense, I will modify it.

src/Utils/Image.php
kukulich: Yes, `is_string` is probably better.

src/Utils/Image.php
kukulich: It's more readable with `round` for me. So it's up to @dg :)

dg: round() is more readable and works with negative numbers, which are used by some methods in class Image. (ie. round(-1.6) === -2).

kukulich:
Forced pushed so it's ready to merge I think.

src/Utils/FileSystem.php
```diff
@@ -78,8 +78,8 @@ public static function delete($path)
 			}
 
 		} elseif (is_dir($path)) {
-			foreach (new \FilesystemIterator($path) as $item) {
-				static::delete($item);
+			foreach (new \FilesystemIterator($path) as $itemPath => $item) {
+				static::delete($itemPath);
```
dg: What about to use `(string) $itemPath`? To save one variable?

kukulich: Or `static::delete($item->getPathname());` ?

dg: Yes.

kukulich: Force pushed.

dg:
Great, thank you!

PR #78, opened 2015-08-23
Trim Character Mask extracted into CONST

Achse:
**BC break**: No
**Doc**: not-needed

**Motivation**: Sometimes I need trim also characters like  `ZERO WIDTH NO-BREAK SPACE` (yes, sadly there is an API that uses it in JSON response). If I want to use `Strings::trim` it forces me to *copy-paste* code for extending the character-mask.

```php
$zeroWithNoBreakSpace =  "\xEF\xBB\xBF";
Strings::trim($rawResponse, " \t\n\r\0\x0B\xC2\xA0" . $zeroWithNoBreakSpace);
```

vs. 

```php
$zeroWithNoBreakSpace =  "\xEF\xBB\xBF";
Strings::trim($rawResponse, Strings::TRIM_CHARACTER_MASK . $zeroWithNoBreakSpace);
```

***

**Note**: *I'm not really sure about naming for that constant. I will be happy if you suggest some better name.*

### DIFF
--- a/src/Utils/Strings.php
+++ b/src/Utils/Strings.php
@@ -16,6 +16,9 @@ use Nette;
 class Strings
 {
 
+	const TRIM_CHARACTERS = " \t\n\r\0\x0B\xC2\xA0";
+
+
 	/**
 	 * Static class - cannot be instantiated.
 	 */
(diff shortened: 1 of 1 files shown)


dg:
Good idea. In fact, it is not mask (unlike argument in trim()), it is just list of characters. What about `TRIM_CHARACTERS`?

Achse:
Sound good. :+1: It's there.

dg:
Please use commit message like `Strings: added const ...`

Achse:
Done.

dg:
Thanks

PR #80, opened 2015-08-23
PHP 5.6: Use variadics and argument unpacking

Majkl578:
Depends on #79 (for composer & Travis PHP version bump).

### DIFF
--- a/src/Utils/Callback.php
+++ b/src/Utils/Callback.php
@@ -45,8 +45,8 @@ class Callback
 
 		self::check($callable);
 		$_callable_ = $callable;
-		return function () use ($_callable_) {
-			return call_user_func_array($_callable_, func_get_args());
+		return function (...$args) use ($_callable_) {
+			return $_callable_(...$args);
 		};
 	}
 
@@ -55,10 +55,10 @@ class Callback
 	 * Invokes callback.
 	 * @return mixed
 	 */
-	public static function invoke($callable)
+	public static function invoke($callable, ...$args)
 	{
 		self::check($callable);
-		return call_user_func_array($callable, array_slice(func_get_args(), 1));
+		return call_user_func_array($callable, $args);
 	}
 
 
@@ -87,13 +87,13 @@ class Callback
 			if ($file === __FILE__ && $onError(str_replace("$function(): ", '', $message), $severity) !== FALSE) {
 				return;
 			} elseif ($prev) {
-				return call_user_func_array($prev, func_get_args());
+				return $prev(...func_get_args());
 			}
 			return FALSE;
 		});
 
 		try {
-			$res = call_user_func_array($function, $args);
+			$res = $function(...$args);
 			restore_error_handler();
 			return $res;
 
--- a/src/Utils/Strings.php
+++ b/src/Utils/Strings.php
@@ -335,10 +335,10 @@ class Strings
 	 * @param  string|array
 	 * @return string
 	 */
-	public static function findPrefix($strings)
+	public static function findPrefix(...$strings)
 	{
-		if (!is_array($strings)) {
-			$strings = func_get_args();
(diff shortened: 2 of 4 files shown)


src/Utils/Strings.php
```diff
@@ -332,10 +332,10 @@ public static function compare($left, $right, $len = NULL)
 	 * @param  string|array
 	 * @return string
 	 */
-	public static function findPrefix($strings)
+	public static function findPrefix(...$strings)
 	{
-		if (!is_array($strings)) {
-			$strings = func_get_args();
+		if (is_array($strings[0])) {
+			$strings = $strings[0];
```
JanTvrdik: This is not so simple because it changes signature. Previously the signature suggested that `findPrefix(['bar', 'baz'])` is preferred. Now it suggests that `findPrefix('bar', 'baz')` is preferred. Do we really want to prefer the variadic syntax over normal array? Passing directly array is slightly faster on PHP 7. AFAIK we should pick one syntax and deprecate the other.

Majkl578: That is true, also the name implies one string argument, not an array. In PHP 7 we should pick one and stick with it (`array $strings` vs `string ...$strings`).
This is a common scenario through Nette though, e.g. nette/finder#4 is similar.
Either way this is a good thing to think about - do we want to prefer variadics & unpacking or untyped arrays?

foxycode: I vote for variadic. Anyone can always unpack array in function call when needed

kukulich: I vote for variadic because of better type hint `string ...$strings`

dg: IMHO scenario `if (!is_array($var)) $var = func_get_args()` can be deprecated when Nette will require >= 5.6.

JanTvrdik: @dg Do the signature would be changed to `array $strings`?

src/Utils/Callback.php
```diff
@@ -55,10 +55,10 @@ public static function closure($callable, $m = NULL)
 	 * Invokes callback.
 	 * @return mixed
 	 */
-	public static function invoke($callable)
+	public static function invoke($callable, ...$args)
 	{
 		self::check($callable);
-		return call_user_func_array($callable, array_slice(func_get_args(), 1));
+		return $callable(...$args);
```
dg: As long as the PHP 5.6 is required, there must stay `call_user_func_array`.

src/Utils/Callback.php
```diff
@@ -69,7 +69,7 @@ public static function invoke($callable)
 	public static function invokeArgs($callable, array $args = [])
 	{
 		self::check($callable);
-		return call_user_func_array($callable, $args);
+		return $callable(...$args);
```
dg: The same here.

src/Utils/Callback.php
```diff
@@ -87,13 +87,13 @@ public static function invokeSafe($function, array $args, $onError)
 			if ($file === __FILE__ && $onError(str_replace("$function(): ", '', $message), $severity) !== FALSE) {
 				return;
 			} elseif ($prev) {
-				return call_user_func_array($prev, func_get_args());
+				return $prev(...func_get_args());
```
dg: Here is `$prev(...)` fine.

src/Utils/Strings.php
dg: This function is used very rarely, but common usecase is probably `findPrefix($a, $b)`, so `...$strings` fits better and posibility to use typehint `string ...$strings` sound good.

Majkl578: :+1: The possibility to typehint the value with variadics is good.

> Passing directly array is slightly faster on PHP 7.

That makes sense, but is the difference noticeable in runtime?

PR #81, opened 2015-08-24
Remove PHP <=5.6 compatibility checks

Majkl578:
Depends on #79.


dg:
static::$messages can be probably removed at all.

Majkl578:
I'd rather keep that for different PR, it involves some changes in exception messages as well and those seem to be a bit crazy (break tests).

PR #82, opened 2015-08-24
Concept: Utils\DateTime deprecation & replacement

Majkl578:
Hi,

this is an attempt to phase out `Nette\Utils\DateTime` and replace it by a new helper class.
The current DateTime makes date & time manipulation easier since it enhances `\DateTime`, yet it has some major flaws: mutability and inheritance.

Why should we want the new helper class?
- current DateTime contains some useful methods to ease manipulation with DateTime instance, these are moved to this new class;
- supports both mutable `\DateTime` and immutable `\DateTimeImmutable` variants, through `\DateTimeInterface`;
- inheritance-free, does not extend mutable `\DateTime`;
- does not provide magical __toString (which by the way drops the time zone information).

This PR is a draft, not a final proposal yet. It consists of new tests for existing DateTIme class as well as the actual replacement with tests as well.

I'd like to hear some feedback -- whether is it really what we want (I hope so) and what & how could be eventually changed.

(This depends on #79 which bumps composer.json & Travis versions. Hence Travis fails.)

### DIFF
--- a/src/Utils/DateHelpers.php
+++ b/src/Utils/DateHelpers.php
@@ -1,0 +1,180 @@
+<?php
+
+namespace Nette\Utils;
+
+use Nette;
+
+/**
+ * DateTime manipulation helpers.
+ */
+class DateHelpers
+{
+	/** minute in seconds */
+	const MINUTE = 60;
+
+	/** hour in seconds */
+	const HOUR = 60 * self::MINUTE;
+
+	/** day in seconds */
+	const DAY = 24 * self::HOUR;
+
+	/** week in seconds */
+	const WEEK = 7 * self::DAY;
+
+	/** average month in seconds */
+	const MONTH = 2629800;
+
+	/** average year in seconds */
+	const YEAR = 31557600;
+
+
+	public function __construct()
+	{
+		throw new Nette\StaticClassException;
+	}
+
+
+	/**
+	 * @param string|int|\DateTimeInterface $time
+	 * @return \DateTime
+	 */
+	public static function createMutable($time)
+	{
+		if ($time instanceof \DateTime) {
+			return clone $time;
+
+		} elseif ($time instanceof \DateTimeImmutable) {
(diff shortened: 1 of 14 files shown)


pavelkouril:
I definitely like the idea. :+1:

fprochazka:
Good job :thumbsup:

dg:
createMutable() vs createImmutable makes users think…

JanTvrdik:
:+1: For immutable only. The only good use-case for mutable DateTime is compatibility with old libraries which accepts `DateTime` only. Unfortunately there is no easy way in PHP core to convert `DateTimeImmutable` to `DateTime`.

Majkl578:
> createMutable() vs createImmutable() makes users think…

Agreed, maybe it could be `create()` vs. `createMutable()` or something. But I'd like to keep both for the reasons below.

> The only good use-case for mutable DateTime is compatibility with old libraries which accepts `DateTime` only.

Exactly. And I think that is very good reason for supporting it. As an example, whole Doctrine stack still does not natively support `DateTimeImmutable`/`DateTimeInterface` (or probably any other library following semver until they bump major version).

> Unfortunately there is no easy way in PHP core to convert `DateTimeImmutable` to `DateTime`.

Actually, I really thought it was in (I used it in this PR, didn't even realize), but it was reverted afterwards, because Derick was against it. :( https://github.com/php/php-src/pull/1145
This is imho yet another reason to have createMutable, which could handle this stupid boilerplate internally.

dg:
Something like this would be useful, but we need to think of a better class name and method names.

enumag:
`create` + `createMutable` seems good enough to me. As for class I'd use `DateTimeHelpers` but that's not important.

dg:
Not in 2.4.0, but it can be added in next minor.

Majkl578:
What to do with this? Still doable for 3.0 where DateTime with inheritance sadly still exists, or should I just close it?

PR #84, opened 2015-08-27
Fixed tests in PHP7 strict mode

kukulich:
1. `Callback::invokeSafe`  - `trim([])` throws `TypeError` in strict mode so I picked another function.

2.  `Image::rotate()` test does't work in strict mode because `imagecolorallocatealpha` returns `false`. The image resource is changed somehow by other tests. It's safer to run all tests independently.

### DIFF
--- a/tests/Utils/Callback.invokeSafe.phpt
+++ b/tests/Utils/Callback.invokeSafe.phpt
@@ -24,23 +24,23 @@ Assert::same('OK1', $res);
 
 
 // skipped error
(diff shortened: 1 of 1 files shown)


dg:
Problem with images is fixed here https://github.com/nette/utils/commit/780f69d1df67a1a8acb130f16df639771d70a38d

kukulich:
Forced push only with the `Callback::invokeSafe` test change.

dg:
:+1:

PR #85, opened 2015-08-28
Strings: Allow to specify separator in webalize()

legendik:
It's helpful if you need remove accents from string and change whitespace separator to something else then `_`.

### DIFF
--- a/src/Utils/Strings.php
+++ b/src/Utils/Strings.php
@@ -196,17 +196,18 @@ class Strings
 	 * Converts to web safe characters [a-z0-9-] text.
 	 * @param  string  UTF-8 encoding
 	 * @param  string  allowed characters
+	 * @param  string  Whitespace separator
 	 * @param  bool
 	 * @return string
 	 */
-	public static function webalize($s, $charlist = NULL, $lower = TRUE)
+	public static function webalize($s, $charlist = NULL, $lower = TRUE, $separator = '-')
 	{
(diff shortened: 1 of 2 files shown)


Budry:
:+1:

fprochazka:
I like it!

src/Utils/Strings.php
```diff
@@ -196,17 +196,18 @@ public static function toAscii($s)
 	 * Converts to web safe characters [a-z0-9-] text.
 	 * @param  string  UTF-8 encoding
 	 * @param  string  allowed characters
+	 * @param  string  Whitespace separator
```
Majkl578: Whitespace?

Budry:
:+1:

dg:
Why?

fprochazka:
@dg because this

```php
$s = Strings::webalize('1 2 3', NULL, TRUE, '_');
```

is nicer than

```php
$s = str_replace('-', '_', Strings::webalize('1 2 3'));
```

dg:
Why it is useful?

JanTvrdik:
The usecase is too narrow.

hrach:
`$s = Strings::webalize('1 2 3', NULL, TRUE, '_');` is totally criptic, mainly because od the two boolean parameters.

dg:
This is edge case and should be done via a different function.

PR #87, opened 2015-09-27
ObjectMixin: warning on method getter in 2.3

dg:


JanTvrdik:
:-1: This must not be done in 2.3. There will most-likely more BC breaks in Nette 3.0 than just this. You cannot add warning for each deprecated feature. It needs more systematic approach.

dg:
Title was wrong: it is not about deprecating, it is about warning.

JanTvrdik:
Does not matter how you call it, if the default value of `warnOnMethodGetter` is `TRUE`, you essentially made the feature deprecated in stable version.

Even if you keep the default value in 2.3 on `FALSE` I would still be strongly against is as I see this as a random, non-systematic approach to much more complex problem than just one magic getter.

dg:
Warning for hidden dangerous bug is useful even in 2.3

pavelkouril:
I agree with @JanTvrdik.

This is a BC break and should not be done. Sure, maybe the feature was a mistake and shouldn't have been added to Nette in the first place, but it's already there, and you just shouldn't remove it like this in a non-major release.

dg:
I am not removing it.

enumag:
:-1: This should not be changed in 2.3 in my opinion.

dg:
Try this solution in your projects and report how it works or suggest different solution. „Be against“ and :-1: is not constructive.

enumag:
I'm not convinced that this needs to be solved (in the stable version).

JanTvrdik:
@dg The solution is simple – don't do this in stable version.

> Warning for hidden dangerous bug

That's quite a stretch. JS (and other languages) has very similar feature since ever and they live just fine.

dg:
But it doesn't exist in PHP. Nette\Object makes PHP more dangerous than is.

JanTvrdik:
Yep and now we have to live with it.

BTW: I would be fine with fix limited to `isLoggedIn()` (and possible few other methods) only.

dg:
That's exactly what I am trying to do. The question is how to define "few other methods".

nechutny:
What about throwing warning only when used as $instance->method and not $this->method ? If is someone's using it in $this context (eg. $form->onSuccess[] = $this->processForm) he should know what it is (property or method).

dg:
@nechutny good idea, but I am not sure whether it could be detected.

nechutny:
@dg: I'm not sure if it is good solution, but It can be implemented. Just move check to __get() resp. get() and increase indexes by one.

	<?php

	class foo {
		public function method()
		{
			$tmp = debug_backtrace(DEBUG_BACKTRACE_PROVIDE_OBJECT | DEBUG_BACKTRACE_IGNORE_ARGS, 2);
			if(isset($tmp[1]) && $tmp[0]['class'] == $tmp[1]['class'])
			{
				echo "called from class\n";
			}
			else
			{
				echo "called from out\n";
			}
		}

		public function ok()
		{
			$this->method();
		}
	}


	$x = new foo();

	$x->method();
	$x->ok();

fprochazka:
How about instead of "breaking" the Object class, let's take it apart and introduce traits without this unwanted behaviour and depecated the entire class?

JanTvrdik:
@fprochazka a) You can't do it in 2.3 b) Yes, Object has a good chance to be replaced with trait in 3.0, hopefully deprecated a bit more stuff (such as the terrible `@method` annotations).

dg:
Info on czech forum  https://forum.nette.org/cs/24419-nette-object-a-reseni-problemu-s-method-gettery

PR #89, opened 2015-10-16
Json::encode - pass all options to json_encode

ondrejmirtes:
Main motivation is to be able to use `JSON_PRESERVE_ZERO_FRACTION` while still benefitting from other features of Json::encode like error detection.

There are no intentional BC breaks involved.

### DIFF
--- a/src/Utils/Json.php
+++ b/src/Utils/Json.php
@@ -16,7 +16,7 @@ use Nette;
 class Json
 {
 	const FORCE_ARRAY = 0b0001;
-	const PRETTY = 0b0010;
+	const PRETTY = JSON_PRETTY_PRINT;
 
 
 	/**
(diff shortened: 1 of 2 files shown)


JanTvrdik:
JSON_PRESERVE_ZERO_FRACTION should be default behavior

ondrejmirtes:
But that would be a big BC break.

dg:
To pass all options goes against the spirit of this class. 

I agree that JSON_PRESERVE_ZERO_FRACTION should be default in 3.0 (due to compatibility and due required minimal version) because stripping zero fraction is bug in fact.

It is no BC break when you exports PHP -> JavaScript, but it is BC break when you exports PHP -> PHP. The question is whether it's a problem or bugfix.

JanTvrdik:
I see JSON_PRESERVE_ZERO_FRACTION  as bugfix. The previous behavior makes not sense. Why would anyone expect and rely on `json_decode(json_encode(2.0)) === 2`?

ondrejmirtes:
It might make no sense, but it's the current behaviour and you can't make any assumptions whether applications rely on it or not.

JanTvrdik:
What? Of course you can make assumptions. Otherwise you will never change anything cause everything change is BC break.

ondrejmirtes:
It's OK to do BC breaks but not in a patch or minor version.

PR #90, opened 2015-11-08
Image: added copyMergeAlpha().

radekdostal:
Merge two true colour images with variable opacity while maintaining alpha transparency of both images.

### DIFF
--- a/src/Utils/Image.php
+++ b/src/Utils/Image.php
@@ -498,6 +498,92 @@ class Image extends Nette\Object
 
 
 	/**
+	 * Merge two true colour images with variable opacity while maintaining alpha transparency of both images.
+	 * @param  Image
+	 * @param  mixed  x-coordinate of destination point
+	 * @param  mixed  y-coordinate of destination point
+	 * @param  mixed  x-coordinate of source point
+	 * @param  mixed  y-coordinate of source point
+	 * @param  mixed  source width
+	 * @param  mixed  source height
+	 * @param  int  opacity 0..100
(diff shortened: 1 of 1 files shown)


JanTvrdik:
:-1: Too much code for too narrow use-case.

radekdostal:
I think it isn't narrow use-case. Merging images by method place() produces black background where two pixels with transparency from each image are covered. This new method copyMergeAlpha() produces correctly image (green map on the top with 50% transparency).

example

enumag:
In that case maybe this method should replace the original place method.

radekdostal:
The dark side of this method is its slowness, because iterating through all pixels. I think, that it has specific use only for images with transparency, so replace the place() method is not a good idea.

enumag:
In that case can you try finding out if the place method can be fixed (without pixel iteration)? I mean there should be same cases of this on the internet.

radekdostal:
I was unable to find a solution without pixel iteration. Moreover, the method signature is not compatible with place() method, because it allows to use even a fragment of the picture.

radekdostal:
It is a [PHP bug](https://bugs.php.net/bug.php?id=23815). There is an opened [pull request](https://github.com/php/php-src/pull/211) with the same algorithm as I took. So I think that only solution is algorithm with pixel iteration.

dg:
This implementation is extreme slow, method has unclear name and a lot of parameters and gives a very different results for opacity = 0 and opacity = 1.

PR #91, opened 2015-11-09
allow to skip online tests

remicollet:
Sometime we need to be able to run test suite from an offline computer (which, at least, is the case in Fedora build system).

The SKIP_ONLINE_TESTS env. variable is quite common (e.g. used in php test suite)

### DIFF
--- a/tests/Utils/FileSystem.phpt
+++ b/tests/Utils/FileSystem.phpt
@@ -41,8 +41,13 @@ test(function () { // copy
 	FileSystem::copy(TEMP_DIR . '/3/file', TEMP_DIR . '/3/x/file');
 	Assert::same('Hello', file_get_contents(TEMP_DIR . '/3/x/file'));
 
-	FileSystem::copy('http://example.com', TEMP_DIR . '/3/x/y/file');
-	Assert::true(is_file(TEMP_DIR . '/3/x/y/file'));
+	if (getenv('SKIP_ONLINE_TESTS')) {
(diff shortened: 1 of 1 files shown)


tests/Utils/FileSystem.phpt
```diff
@@ -41,8 +41,13 @@ test(function () { // copy
...
 
-	FileSystem::copy('http://example.com', TEMP_DIR . '/3/x/y/file');
-	Assert::true(is_file(TEMP_DIR . '/3/x/y/file'));
+	if (getenv('SKIP_ONLINE_TESTS') ?: false) {
+		FileSystem::copy(__FILE__, TEMP_DIR . '/3/x/y/file');
+		Assert::true(is_file(TEMP_DIR . '/3/x/y/file'));
+	} else {
+		FileSystem::copy('http://example.com', TEMP_DIR . '/3/x/y/file');
+		Assert::true(is_file(TEMP_DIR . '/3/x/y/file'));
+	}
```
JanTvrdik: Why not just

~~~php
if (!getenv('SKIP_ONLINE_TESTS')) {
    FileSystem::copy('http://example.com', TEMP_DIR . '/3/x/y/file');
    Assert::true(is_file(TEMP_DIR . '/3/x/y/file'));
}
~~~
?

remicollet: else is needed because '/3/x/y/file' is expected to exists in file "copy" test.

milo:
Alternative implementation #93

dg:
Should be closed by #93

PR #92, opened 2015-11-11
Image::place preserves alpha channel

dg:
Related to https://github.com/nette/utils/pull/90


radekdostal:
@dg Your solution is amazing!
For example: two images 620x385px
With 4fcb6e777fc095714cb63fcbf047a750fddcaa51 my script runs about one second.
With abfcf392b2f3f64c7c6dacb57857e7064ff77d0e even just half a second.
(With #90 the script runs about 17 seconds).

enumag:
@dg Awesome!

dg:
It is still order of magnitudes slower than imagecopymerge for images without alpha. And I don't know how to detect whether alpha is used or not.

radekdostal:
Maybe http://php.net/manual/en/function.getimagesize.php#105033.

dg:
It is not about files, but about GD images in memory.

radekdostal:
What about php://memory wrapper?
In [this code](https://code.google.com/p/inspired-social/source/browse/trunk/plugin/wp-content/plugins/article2pdf/contributed/fpdf/fpdf_alpha.php?r=10#202) exists method _parsepng() which parses PNG header.

PR #93, opened 2015-11-11
tests: added RemoteStream to be able test non-local URLs offline

milo:


milo:
@dg Not sure you know that [this `@`](https://github.com/nette/utils/blob/d04f1136148e805f609d99ac7e9493668b36a00d/src/Utils/FileSystem.php#L60) mutes:
```
E_WARNING: stream_copy_to_stream(): RemoteStream::stream_stat is not implemented!
E_WARNING: stream_copy_to_stream(): RemoteStream::stream_read is not implemented!
E_WARNING: stream_copy_to_stream(): RemoteStream::stream_eof is not implemented! Assuming EOF
```
but `stream_copy_to_stream()` return `0` after that.

dg:
Error system in PHP is a tragedy…

I do not know how to distinguish between errors that lead to failure from those that are just informative. I can either throw an exception for any error and do not wait for the return value like [here](https://github.com/nette/http/blob/dea87d08590689854e2bcc920d934138eaab80c2/src/Http/Session.php#L89) or throw an exception check on failing return values and remove `@`, so catched exception will just trigger notices and warnings.

PR #95, opened 2015-11-14
Strings::toAscii() optimization

kukulich:
It's expensive to create `Transliterator` instance everytime.

### DIFF
--- a/src/Utils/Strings.php
+++ b/src/Utils/Strings.php
@@ -162,13 +162,18 @@ class Strings
 	 */
 	public static function toAscii($s)
 	{
+		static $transliterator = NULL;
+		if ($transliterator === NULL && class_exists('Transliterator', FALSE)) {
+			$transliterator = \Transliterator::create('Any-Latin; Latin-ASCII');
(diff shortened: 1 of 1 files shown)


src/Utils/Strings.php
```diff
@@ -162,13 +162,18 @@ public static function normalizeNewLines($s)
 	 */
 	public static function toAscii($s)
 	{
+		static $transliterator = null;
```
dg: Please `NULL`.

src/Utils/Strings.php
```diff
@@ -162,13 +162,18 @@ public static function normalizeNewLines($s)
...
+		if ($transliterator === null && class_exists('Transliterator')) {
```
Majkl578: Hmm, any way to also get rid of this repetitive class_exists check?

dg: maybe `class_exists(...., FALSE)`, it is cheap.

kukulich: Done.

PR #96, opened 2015-11-16
Validator::is: fixed value overwriting

matej21:


src/Utils/Validators.php
```diff
@@ -137,14 +137,15 @@ public static function is($value, $expected)
 			}
 
 			if (isset($item[1])) {
+				$rangeValue = $value;
```
dg: Because it is about length (of string, array, etc), what about `$length`?

matej21: updated

PR #99, opened 2015-12-03
ArrayList: Possibility to prepend a value to the start of the internal array

doublemcz:


dg:
Closed by https://github.com/nette/utils/pull/99

PR #102, opened 2016-03-17
FileSystem supports deleting over \Traversable|array

CZechBoY:


fprochazka:
:+1:

dg:
:-1:

JanTvrdik:
* It is inconsistent – why support Finder in delete and not in `copy` and `rename`?
* It is inconsistent – why support Finder and not arbitrary Traversable or array?
* It creates new (soft) dependency – nette/utils are currently unaware of nette/finder

CZechBoY:
I have no idea how to implement rename for collection of files.
Maybe for move.

Dependency on class which is in same package is not problem... I can add class_exists check.

dg:
What is your intention? What this should do?

CZechBoY:
Aaah I see that Finder is in another package! Sorry.
What about Traversable|array as @JanTvrdik mentioned?

fprochazka:
@JanTvrdik having it handle `array|\Traversable` makes a lot more sense :+1:

fprochazka:
> It is inconsistent – why support Finder in delete and not in copy and rename?

What would copy & rename variant do? how would you define the behavior?

src/Utils/FileSystem.php
```diff
@@ -84,6 +84,14 @@ public static function delete($path)
 			if (!@rmdir($path)) { // @ is escalated to exception
 				throw new Nette\IOException("Unable to delete directory '$path'.");
 			}
+		} elseif ($path instanceof \Nette\Utils\Finder) {
+			/** @var \SplFileInfo $item */
+			foreach ($path as $item) {
+				static::delete($item->getRealPath());
+			}
+			if (!@rmdir($item->getPath())) { // @ is escalated to exception
```
dg: What this mean?

CZechBoY: Delete directory where the last file was.
Probably I can't use Finder for finding in more than one directory for example:
`
$finder->find('images/*/*.jpg');
`

dg: What is the last file? It is completely random.

CZechBoY: Yes, if I can iterate over more subdirectories.
Suddenly I didn't find out how to search over subdirectories.

dg: Do you agree that function that deletes some files and than deletes directory of one random file is not needed by core utils? It is very special use case.

JanTvrdik:
> What would copy & rename variant do?

The same as CLI unix tools, e.g. 

* `FileSystem::move(Finder::find('*.php'), 'some_directory')` = `mv *.php some_directory`
* `FileSystem::copy(Finder::find('*.php'), 'some_directory')` = `cp *.php some_directory`
* `FileSystem::delete(Finder::find('*.php'))` = `rm *.php`

But I'm not convinces that even supporting general Traversable|array consistently in all of those functions is good idea, because

* You need to think of all possible behavior edge-cases and handle them well.
* When you need to do it, you can just as well use `foreach`.

src/Utils/FileSystem.php
CZechBoY: I wanted to delete all selected files and the directory in which they are. There is missing condition on empty directory which should be removed. And second problem is that files can be in different directories - I dont know if it is possible with Finder.

JanTvrdik: > I wanted to delete all selected files and the directory in which they are

Why not just delete the directory then?

CZechBoY: Because I don't know if I will erase whole directory or not. There is missing check for it.

dg:
FileSystem::delete should stay simple.

PR #105, opened 2016-04-04
Nette\SmartObject & StaticClass

dg:
https://forum.nette.org/cs/25958-jak-se-vyporadat-s-nette-object (Czech)

### DIFF
--- a/src/Utils/ObjectMixin.php
+++ b/src/Utils/ObjectMixin.php
@@ -270,6 +270,57 @@ class ObjectMixin
 	}
 
 
+	/********************* magic @properties ****************d*g**/
+
+
+	/**
+	 * Returns array of magic properties defined by annotation @property.
+	 * @return array of [name => bit mask]
+	 */
+	public static function getMagicProperties($class)
+	{
+		static $cache;
+		$props = & $cache[$class];
+		if ($props !== NULL) {
+			return $props;
+		}
+
+		$rc = new \ReflectionClass($class);
+		preg_match_all(
+			'~^  [ \t*]*  @property(|-read|-write)  [ \t]+  [^\s$]+  [ \t]+  \$  (\w+)  ()~mx',
+			(string) $rc->getDocComment(), $matches, PREG_SET_ORDER
+		);
+
+		$props = [];
+		foreach ($matches as list(, $type, $name)) {
+			$uname = ucfirst($name);
+			$write = $type !== '-read'
+				&& $rc->hasMethod($nm = 'set' . $uname)
+				&& ($rm = $rc->getMethod($nm)) && $rm->getName() === $nm && !$rm->isPrivate() && !$rm->isStatic();
+			$read = $type !== '-write'
+				&& ($rc->hasMethod($nm = 'get' . $uname) || $rc->hasMethod($nm = 'is' . $uname))
+				&& ($rm = $rc->getMethod($nm)) && $rm->getName() === $nm && !$rm->isPrivate() && !$rm->isStatic();
+
+			if ($read || $write) {
+				$props[$name] = $read << 0 | ($nm[0] === 'g') << 1 | $rm->returnsReference() << 2 | $write << 3;
+			}
+		}
+
+		if ($parent = get_parent_class($class)) {
+			$props += self::getMagicProperties($parent);
+		}
+		return $props;
+	}
+
+
+	/** @internal */
+	public static function getMagicProperty($class, $name)
+	{
+		$props = self::getMagicProperties($class);
+		return isset($props[$name]) ? $props[$name] : NULL;
+	}
+
+
 	/********************* magic @methods ****************d*g**/
 
 
--- a/src/Utils/ObjectMixin.php
+++ b/src/Utils/ObjectMixin.php
@@ -48,9 +48,6 @@ class ObjectMixin
 		if ($name === '') {
 			throw new MemberAccessException("Call to class '$class' method without name.");
 
-		} elseif ($isProp && $_this->$name instanceof \Closure) { // closure in property
-			return call_user_func_array($_this->$name, $args);
-
 		} elseif ($isProp === 'event') { // calling event handlers
 			if (is_array($_this->$name) || $_this->$name instanceof \Traversable) {
 				foreach ($_this->$name as $handler) {
@@ -59,6 +56,9 @@ class ObjectMixin
 			} elseif ($_this->$name !== NULL) {
 				throw new Nette\UnexpectedValueException("Property $class::$$name must be array or NULL, " . gettype($_this->$name) . ' given.');
 			}
+
+		} elseif ($isProp && $_this->$name instanceof \Closure) { // closure in property
+			return call_user_func_array($_this->$name, $args);
 
 		} elseif (($methods = & self::getMethods($class)) && isset($methods[$name]) && is_array($methods[$name])) { // magic @methods
 			list($op, $rp, $type) = $methods[$name];
--- a/src/Utils/ObjectMixin.php
+++ b/src/Utils/ObjectMixin.php
@@ -16,13 +16,7 @@ use Nette\MemberAccessException;
  */
 class ObjectMixin
 {
-	/** @var array (name => 0 | bool | array)  used by getMethods() */
-	private static $methods = [];
-
-	/** @var array (name => 'event' | TRUE)  used by hasProperty() */
-	private static $props = [];
-
-	/** @var array (name => [type => callback])  used by get|setExtensionMethod() */
+	/** @var array [name => [type => callback]] used by extension methods */
 	private static $extMethods = [];
 
 
@@ -33,6 +27,9 @@ class ObjectMixin
 	{
 		throw new Nette\StaticClassException;
 	}
+
+
+	/********************* Nette\Object ****************d*g**/
 
 
 	/**
@@ -84,8 +81,7 @@ class ObjectMixin
 			return $_this;
 
 		} elseif ($cb = self::getExtensionMethod($class, $name)) { // extension methods
-			array_unshift($args, $_this);
-			return Callback::invokeArgs($cb, $args);
+			return Callback::invoke($cb, $_this, ...$args);
 
 		} else {
 			$hint = self::getSuggestion(array_merge(
@@ -240,41 +236,7 @@ class ObjectMixin
 	}
 
 
-	/**
-	 * Checks if the public non-static property exists.
-	 * @return mixed
-	 */
-	private static function hasProperty($class, $name)
-	{
-		$prop = & self::$props[$class][$name];
-		if ($prop === NULL) {
-			$prop = FALSE;
-			try {
-				$rp = new \ReflectionProperty($class, $name);
-				if ($rp->isPublic() && !$rp->isStatic()) {
-					$prop = $name >= 'onA' && $name < 'on_' ? 'event' : TRUE;
-				}
-			} catch (\ReflectionException $e) {
-			}
-		}
-		return $prop;
-	}
-
-
-	/**
-	 * Returns array of public (static, non-static and magic) methods.
-	 * @return array
-	 */
-	private static function & getMethods($class)
-	{
-		if (!isset(self::$methods[$class])) {
-			self::$methods[$class] = array_fill_keys(get_class_methods($class), 0) + self::getMagicMethods($class);
-			if ($parent = get_parent_class($class)) {
-				self::$methods[$class] += self::getMethods($parent);
-			}
-		}
-		return self::$methods[$class];
-	}
+	/********************* magic @methods ****************d*g**/
 
 
 	/**
@@ -383,6 +345,9 @@ class ObjectMixin
 	}
 
 
+	/********************* extension methods ****************d*g**/
+
+
 	/**
 	 * Adds a method to class.
 	 * @param  string
@@ -438,6 +403,9 @@ class ObjectMixin
 	}
 
 
+	/********************* utilities ****************d*g**/
+
+
 	/**
 	 * Finds the best suggestion (for 8-bit encoding).
 	 * @return string|NULL
@@ -468,6 +436,45 @@ class ObjectMixin
 			$doc[] = $rc->getDocComment();
 		} while ($rc = $rc->getParentClass());
 		return preg_match_all($pattern, implode($doc), $m) ? $m[1] : [];
+	}
+
+
+	/**
+	 * Checks if the public non-static property exists.
+	 * @return bool|'event'
+	 */
+	private static function hasProperty($class, $name)
+	{
+		static $cache;
+		$prop = & $cache[$class][$name];
+		if ($prop === NULL) {
+			$prop = FALSE;
+			try {
+				$rp = new \ReflectionProperty($class, $name);
+				if ($rp->isPublic() && !$rp->isStatic()) {
+					$prop = $name >= 'onA' && $name < 'on_' ? 'event' : TRUE;
+				}
+			} catch (\ReflectionException $e) {
+			}
+		}
+		return $prop;
+	}
+
+
+	/**
+	 * Returns array of public (static, non-static and magic) methods.
+	 * @return array
+	 */
+	private static function & getMethods($class)
+	{
+		static $cache;
+		if (!isset($cache[$class])) {
+			$cache[$class] = array_fill_keys(get_class_methods($class), 0) + self::getMagicMethods($class);
+			if ($parent = get_parent_class($class)) {
+				$cache[$class] += self::getMethods($parent);
+			}
+		}
+		return $cache[$class];
 	}
 
 }
--- a/src/Utils/ObjectMixin.php
+++ b/src/Utils/ObjectMixin.php
@@ -63,12 +63,12 @@ class ObjectMixin
 	/**
 	 * @throws MemberAccessException
 	 */
-	public static function strictCall($class, $method)
+	public static function strictCall($class, $method, $additionalMethods = [])
 	{
 		$hint = self::getSuggestion(array_merge(
 			get_class_methods($class),
 			self::parseFullDoc(new \ReflectionClass($class), '~^[ \t*]*@method[ \t]+(?:\S+[ \t]+)??(\w+)\(~m'),
-			array_keys(self::getExtensionMethods($class))
+			$additionalMethods
 		), $method);
 
 		if (method_exists($class, $method)) { // called parent::$method()
@@ -146,7 +146,7 @@ class ObjectMixin
 			return Callback::invoke($cb, $_this, ...$args);
 
 		} else {
-			self::strictCall($class, $name);
+			self::strictCall($class, $name, array_keys(self::getExtensionMethods($class)));
 		}
 	}
 
--- a/src/Utils/ObjectMixin.php
+++ b/src/Utils/ObjectMixin.php
@@ -29,6 +29,66 @@ class ObjectMixin
 	}
 
 
+	/********************* strictness ****************d*g**/
+
+
+	/**
+	 * @throws MemberAccessException
+	 */
+	public static function strictGet($class, $name)
+	{
+		$hint = self::getSuggestion(array_merge(
+			array_keys(get_class_vars($class)),
+			self::parseFullDoc($class, '~^[ \t*]*@property(?:-read)?[ \t]+(?:\S+[ \t]+)??\$(\w+)~m')
+		), $name);
+		throw new MemberAccessException("Cannot read an undeclared property $class::\$$name" . ($hint ? ", did you mean \$$hint?" : '.'));
+	}
+
+
+	/**
+	 * @throws MemberAccessException
+	 */
+	public static function strictSet($class, $name)
+	{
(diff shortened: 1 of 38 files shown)


src/Utils/Strict.php
```diff
@@ -0,0 +1,90 @@
...
+		$items = array_diff($rc->getProperties(\ReflectionProperty::IS_PUBLIC), $rc->getProperties(\ReflectionProperty::IS_STATIC));
+		$hint = ($t = ObjectMixin::getSuggestion($items, $name)) ? ", did you mean $$t?" : '.';
+		throw new MemberAccessException("Attempt to write to undeclared property {$rc->getName()}::$$name$hint");
+	}
+
+
+	/**
+	 * @return bool
+	 */
+	public function __isset($name)
```
fprochazka: I was thinking about this earlier when I was [making this](https://github.com/Kdyby/StrictObjects/blob/b334bea700510844f6de31d2c2f6030f9ea3c26d/src/Scream.php#L89-L101) and I was unable to come up with benefit of returning `FALSE` here.
In Latte, it might make sense, since there are magic properties on the template object, but it does not make sense generaly imho.

If you're beeing trully strict, this potentially hides the fact, that you're doing an `isset()` check on property that is not defined.
Also, you probably never wanna do isset checks on classes that are strict, since you simply expect the property be in the definition.

dg: You are right that `return FALSE` is useless a method can be removed. I'll remove it.

src/Utils/Strict.php
```diff
@@ -0,0 +1,81 @@
...
+{
+
+	/**
+	 * Call to undefined method.
+	 * @throws MemberAccessException
+	 */
+	public function __call($name, $args)
+	{
+		$class = method_exists($this, $name) ? 'parent' : get_class($this);
+		$items = (new \ReflectionClass($this))->getMethods(\ReflectionMethod::IS_PUBLIC);
```
fprochazka: I'm aware that you can call static members as methods, but should they be suggested?

dg: Why not? You can make typo in static method called as non-static.

fprochazka: Well, that is my point. Should the static methods be suggested? Since they're static and there is no reason to call them as non-static.

src/Utils/StrictPlus.php
```diff
@@ -0,0 +1,233 @@
...
+
+/**
+ * Strict class for better experience.
+ * - 'did you mean' hints
+ * - access to undeclared members throws exceptions
+ * - support for @property annotations
+ * - support for calling event handlers stored in $onEvent via onEvent()
+ * - compatible with Nette\Object
+ */
+trait StrictPlus
```
Majkl578: Can anyone think of better name for this? The _plus_part implies even more strict behavior, but it's not the case, instead it adds totally non-strict syntactic sugar...

Majkl578:
:+1: I wish this was made a separate repo/project.

src/Utils/Strict.php
```diff
@@ -0,0 +1,88 @@
...
+		), $name);
+		throw new MemberAccessException("Call to undefined static method {$rc->getName()}::$name()" . ($hint ? ", did you mean $hint()?" : '.'));
+	}
+
+
+	/**
+	 * Access to undeclared property.
+	 * @throws MemberAccessException
+	 */
+	public function &__get($name)
```
Majkl578: No need for reference here.

src/Utils/StrictPlus.php
```diff
@@ -0,0 +1,233 @@
...
+	/**
+	 * @return void
+	 * @throws MemberAccessException
+	 */
+	public function __unset($name)
+	{
+		$class = get_class($this);
+		if (!ObjectMixin::hasProperty($class, $name)) { // strict class
+			throw new MemberAccessException("Cannot unset the property $class::\$$name.");
+		}
```
Majkl578: This seems to pretend that property was sucessfully unset, but it was not, getter on such property would still work, wouldn't it?

src/Utils/StrictPlus.php
dg: StrictSugar?

src/Utils/StrictPlus.php
dg: Getter will fail https://github.com/nette/utils/blob/cf3bb5e487e57776103828d6b5ad2a446fcdf17d/tests/Utils/Object.unsetProperty.phpt#L40-L43

src/Utils/StrictPlus.php
dg: …or to not complicate it and have only one trait…

matej21: BlackMagic? :P

JanTvrdik: Enchantments? StrictEnhanced? StrictButComfortable?

fprochazka: Enchantments :+1:

pavelkouril: Enchantments sounds like something from a fantasy game and isn't really describing what it does. :(

JanTvrdik: > **Enchantments** sounds like something from a fantasy game and isn't really describing what it does. 

I think it sort-of does. It casts a couple of spells on the given class to alter it's behavior. It's similar to BlackMagic but sounds more optimistic.

src/Utils/ObjectMixin.php
```diff
@@ -317,6 +306,45 @@ public static function getMagicMethods($class)
...
+		foreach ($matches as list(, $type, $name)) {
+			$uname = ucfirst($name);
+			$write = (!$type || $type === '-write')
+				&& $rc->hasMethod($nm = 'set' . $uname)
+				&& ($rm = $rc->getMethod($nm)) && $rm->getName() === $nm && !$rm->isPrivate() && !$rm->isStatic();
+			$read = (!$type || $type === '-read')
+				&& ($rc->hasMethod($nm = 'get' . $uname) || $rc->hasMethod($nm = 'is' . $uname))
+				&& ($rm = $rc->getMethod($nm)) && $rm->getName() === $nm && !$rm->isPrivate() && !$rm->isStatic();
+			if ($read || $write) {
+				$props[$name] = [$read ? $nm : FALSE, $write, $read ? $rm->returnsReference() : NULL];
```
JanTvrdik: * Why store method name for getter and only boolean for setter?
*  `$read ? $rm->returnsReference() : NULL` could be simplified to `$read && $rm->returnsReference()`

JanTvrdik: If we would store only boolean for getter as well, we may replace the 2-level array with a single integer (used as bit-flags storage) for each property, which should result in lower memory footprint and faster code.

~~~php
$ref = $read && $rm->returnsReference();
$props[$name] = ($read << 0) | ($write << 1) || ($red << 2);
~~~

src/Utils/ObjectMixin.php
```diff
@@ -25,14 +27,8 @@ class ObjectMixin
...
-
-	/**
-	 * Static class - cannot be instantiated.
-	 */
-	final public function __construct()
-	{
-		throw new Nette\StaticClassException;
-	}
+	/** @var array (name => [read name, write?, returnsByRef?])  used by getMagicProperties() */
+	private static $properties = [];
```
JanTvrdik: Having both `$props` and `$properties` is bad.

src/Utils/ObjectMixin.php
```diff
@@ -317,6 +306,45 @@ public static function getMagicMethods($class)
 
 
 	/**
+	 * Returns array of magic properties defined by annotation @property.
+	 * @return array
+	 */
+	public static function getMagicProperties($class)
```
JanTvrdik: But on how this method is actually used, having `getMagicProperty($class, $name)` (likely returning either NULL or 0 if such property does not exist) may be reasonable.

src/Utils/ObjectMixin.php
dg: Setter is always `setAbc`, getter is `getAbc` or `isAbc`.

src/Utils/ObjectMixin.php
dg: Yes, but $props will be removed in future.

src/Utils/StrictPlus.php
dg: Comfortable sound good.

src/Utils/StrictPlus.php
```diff
@@ -0,0 +1,244 @@
...
+		$isProp = ObjectMixin::hasProperty($class, $name);
+
+		if ($name === '') {
+			throw new MemberAccessException("Call to class '$class' method without name.");
+
+		} elseif ($isProp && $this->$name instanceof \Closure) { // closure in property
+			trigger_error("Invoking closure in property via \$obj->$name() is deprecated" . ObjectMixin::getSource(), E_USER_DEPRECATED);
+			return call_user_func_array($this->$name, $args);
+
+		} elseif ($isProp === 'event') { // calling event handlers
```
JanTvrdik: Is "calling event handlers" intentionally handled after "closure in property"? Is the following behavior intentional?

~~~php
$obj->onUpdate = function () { ... };
$obj->onUpdate() // this will call the closure instead of throwing error that property `$onUpdate` must be "array or NULL".
~~~

dg: Good point. This is primarily problem in Nette\Object, because StrictPlus will throw warning, although with a different message.

src/Utils/ObjectMixin.php
JanTvrdik: Good point, that would require another bit flag. The `__get` initial calling sequence would then look as

~~~php
$class = get_class($this);
$uname = ucfirst($name);
$prop = ObjectMixin::getMagicProperty($class, $name);

if ($name === '') {
    throw new MemberAccessException("Cannot read a class '$class' property without name.");

} elseif ($prop & BIT_MAGIC_A) {
    return $this->{'get' . $uname}();

} elseif ($prop & BIT_MAGIC_B) {
    return $this->{'is' . $uname}();

} elseif ($prop & BIT_MAGIC_C) {
    $value = $this->{'get' . $uname}();
    return $value;

} elseif ($prop & BIT_MAGIC_D) {
    $value = $this->{'is' . $uname}();
    return $value;
}
~~~

Yeah, I know, over-optimization but that's just how I like it =)

src/Utils/ObjectMixin.php
JanTvrdik: How would you implement events without `$props`? You can rename in the future the property to `$events` but you still need to store somewhere the information from reflection check.

src/Utils/ObjectMixin.php
```diff
@@ -317,6 +306,45 @@ public static function getMagicMethods($class)
...
+			\$ (\w+)
+		()~mx', (string) $rc->getDocComment(), $matches, PREG_SET_ORDER);
+
+		$props = [];
+		foreach ($matches as list(, $type, $name)) {
+			$uname = ucfirst($name);
+			$write = (!$type || $type === '-write')
+				&& $rc->hasMethod($nm = 'set' . $uname)
+				&& ($rm = $rc->getMethod($nm)) && $rm->getName() === $nm && !$rm->isPrivate() && !$rm->isStatic();
+			$read = (!$type || $type === '-read')
```
JanTvrdik: could be simplified to `$type !== '-write'`

src/Utils/ObjectMixin.php
```diff
@@ -317,6 +306,45 @@ public static function getMagicMethods($class)
...
+		preg_match_all('~^
+			[ \t*]*  @property(|-read|-write)  [ \t]+
+			(?: [^\s]+  [ \t]+ )?
+			\$ (\w+)
+		()~mx', (string) $rc->getDocComment(), $matches, PREG_SET_ORDER);
+
+		$props = [];
+		foreach ($matches as list(, $type, $name)) {
+			$uname = ucfirst($name);
+			$write = (!$type || $type === '-write')
```
JanTvrdik: could be simplified to `$type !== '-read'`

src/Utils/ObjectMixin.php
dg: :+1:

src/Utils/ObjectMixin.php
dg: It means a lot of conditions for every property usage.…

src/Utils/ObjectMixin.php
```diff
@@ -317,6 +306,45 @@ public static function getMagicMethods($class)
...
+				&& $rc->hasMethod($nm = 'set' . $uname)
+				&& ($rm = $rc->getMethod($nm)) && $rm->getName() === $nm && !$rm->isPrivate() && !$rm->isStatic();
```
JanTvrdik: Do you intentionally return accept protected methods as well?

dg: Yes, it is feature.

src/Utils/ObjectMixin.php
JanTvrdik: This first only will almost always match (90+ % by guess), the second one will handle the most of the remaining cases (7 % by guess). Getters returning references (e.g. `& getName()` or `& isValid()`) accessed as properties  are extremely rare.

src/Utils/ObjectMixin.php
dg: I'll probably change it to static variable.

src/Utils/ObjectMixin.php
dg: I tried it here https://github.com/nette/utils/pull/105/commits/c369f2269adda3d8c6eec4a59ff1dc49b0b01aa5

src/Utils/StrictPlus.php
```diff
@@ -0,0 +1,126 @@
...
+				throw new MemberAccessException("Cannot read a write-only property $class::\$$name.");
+			}
+			$m = ($prop & 0b0010 ? 'get' : 'is') . $name;
+			if ($prop & 0b0100) {
+				return $this->$m();
+			} else {
+				$val = $this->$m();
+				return $val;
+			}
+		} else {
```
JanTvrdik: What about

~~~php
$class = get_class($this);
$prop = ObjectMixin::getMagicProperties($class)[$name] ?? NULL;

if ($prop & 0b0001) {
    $m = ($prop & 0b0010 ? 'get' : 'is') . $name;
    if ($prop & 0b0100) {
        return $this->$m();
    } else {
        $val = $this->$m();
        return $val;
    }

} elseif ($prop !== NULL) {
    throw new MemberAccessException("Cannot read a write-only property $class::\$$name.");

} else {
    ObjectMixin::strictGet($class, $name);
}
~~~

src/Utils/ObjectMixin.php
```diff
@@ -461,13 +502,64 @@ public static function getSuggestion(array $items, $value)
...
+		return $prop;
+	}
+
+
+	/**
+	 * Returns array of public (static, non-static and magic) methods.
+	 * @return array
+	 * @internal
+	 */
+	public static function & getMethods($class)
```
JanTvrdik: Looks like this could be private.

dg: This https://github.com/nette/utils/pull/105/commits/d465062927b7c5cc2a71d0131572338caa125fc5 will not be merged, it is only for idea.

JanTvrdik: I know but it's difficult to comment on code just before this commit.

dg: Ok, removed

src/Utils/Strict.php
JanTvrdik: btw: Nette has inconsistent coding style of writing `&` before methods names, it would be nice to unify it.

tests/Utils/ObjectMixin.getMagicProperites().phpt
```diff
@@ -0,0 +1,117 @@
...
+ * @property $cAse
+ * @property $private
+ * @property $protected
+ * @property $static
+ * @property-read $read
+ * @property-write $write
+ * @property-x $invalid
+ * abc @property $invalid2
+ */
+class TestClass
```
JanTvrdik: * You test property annotations without types (except for one case), even though in the wild types are almost always present. 
* We may even consider not supporting property annotations without types (`mixed` can be used if type is unknown) to encourage programmers to write the annotations properly.
* You're missing edge case when `$this` is used as a type.

JanTvrdik: * You test property annotations without types (except for one case), even though in the wild types are almost always present. 
* We may even consider not supporting property annotations without types (`mixed` can be used if type is unknown) to encourage programmers to write the annotations properly.
* You're missing edge case when `$this` is used as a type.

src/Utils/StrictPlus.php
```diff
@@ -0,0 +1,232 @@
...
+				$rp->setValue($this, $args[0]);
+			} elseif ($op === 'add') {
+				$val = $rp->getValue($this);
+				$val[] = $args[0];
+				$rp->setValue($this, $val);
+			}
+			return $this;
+
+		} elseif ($cb = ObjectMixin::getExtensionMethod($class, $name)) { // extension methods
+			trigger_error("Extension methods like $class::$name() are deprecated" . ObjectMixin::getSource(), E_USER_DEPRECATED);
```
JanTvrdik: I think this should say "Extension methods such as" instead of "Extension methods like".

src/Utils/StrictPlus.php
```diff
@@ -0,0 +1,232 @@
...
+	 */
+	public static function extensionMethod($name, $callback = NULL)
+	{
+		if (strpos($name, '::') === FALSE) {
+			$class = get_called_class();
+		} else {
+			list($class, $name) = explode('::', $name);
+			$class = (new \ReflectionClass($class))->getName();
+		}
+		trigger_error("Extension methods like $class::$name() are deprecated" . ObjectMixin::getSource(), E_USER_DEPRECATED);
```
JanTvrdik: I think this should say "Extension methods such as" instead of "Extension methods like".

JanTvrdik: I think this should say "Extension methods such as" instead of "Extension methods like".

tests/Utils/ObjectMixin.getMagicProperites().phpt
dg: ad 1) one test is enough, isn't it?
ad 2) hmmm, yes, it can be loosen always
ad 3) I dont like to support this

src/Utils/StrictPlus.php
dg: :+1:

src/Utils/ObjectMixin.php
```diff
@@ -271,6 +271,58 @@ public static function has($_this, $name)
 	}
 
 
+	/********************* magic @properties ****************d*g**/
+
+
+	/**
+	 * Returns array of magic properties defined by annotation @property.
+	 * @return array of [name => [getter name?, allow write?, returns by ref?]]
```
JanTvrdik: phpDoc not updated

src/Utils/ObjectMixin.php
```diff
@@ -270,6 +263,57 @@ public static function has($_this, $name)
...
+		}
+		return $props;
+	}
+
+
+	/** @internal */
+	public static function getMagicProperty($class, $name)
+	{
+		$props = self::getMagicProperties($class);
+		return isset($props[$name]) ? $props[$name] : NULL;
```
JanTvrdik: I still think that 0 is slightly better than NULL (you can declare return type as int, you avoid casting NULL to integer when using bit-wise and to check some bit, it is shorter)

hrach:
I wonder, wouldn't be better to push traits into Utils subnamespace? Now it's the ideal time.

enumag:
More like into a separate package in my opinion.

dg:
I do not want to force all users Nette and authors of addons to require a new package.

hrach:
I understand that. That's why I suggest putting this into correct package namespace. :)

dg:
I am rather surprised that nobody did not address namespace :-)  https://forum.nette.org/cs/25958-jak-se-vyporadat-s-nette-object#p172395

I have no idea.

JanTvrdik:
> wouldn't be better to push traits into Utils subnamespace?

Good point. I agree, now is the ideal time to do it.

> More like into a separate package in my opinion.

I disagree, nette/utils is imho the ideal place.

> I do not want to force all users Nette and authors of addons to require a new package.

nette/utils would depend on the new package, so no manual change would be required by Nette users

dg:
> nette/utils would depend on the new package, so no manual change would be required by Nette users

Not so easy, because *the new package* depends on nette\utils too and I dont want to think two days how to solve chicken egg problem.

Majkl578:
It's sad to see the wishes to have a StrictObject trait were ignored... Many people were hoping to have a trait with strictness, but without enforced magic. :disappointed:

fprochazka:
Cough, cough, [already exists](https://github.com/Kdyby/StrictObjects), cough, cough :smirk:

castamir:
@fprochazka cool trait, terrible name :D

dg:
StrictObject is not needed.

PR #106, opened 2016-04-07
Allow pass flag = 0 to json_encode on PHP5.4+

kravcik:
```
$array = ['tag' => 'Odómetro'];

\Nette\Utils\Json::encode($array); //"{"Tag":"Odómetro"}"
```

Needed behaviour:

`\Nette\Utils\Json::encode($array, \Nette\Utils\Json::UGLY); //"{"Tag":"Od\u00f3metro"}"`

### DIFF
--- a/src/Utils/Json.php
+++ b/src/Utils/Json.php
@@ -17,7 +17,7 @@ class Json
 {
 	const FORCE_ARRAY = 0b0001;
 	const PRETTY = 0b0010;
-
+	const UGLY = 0b0011;
 
 	/**
 	 * Static class - cannot be instantiated.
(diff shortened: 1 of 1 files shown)


JanTvrdik:
Why?

kravcik:
Most of  Insurance companies using some old API and chars like 'óéíářžš' causes some troubles.

My point is that Nette disable legit option for PHP function for no reason. And its impossible using Utils if PHP version has defined JSON_UNESCAPED_SLASHES or JSON_UNESCAPED_UNICODE.

dg:
For edge cases is probably best solution to use json_encode() directly (and check json_last_error()).

kravcik:
Yeah, im using extend version of Utils.

Just wanna suggest more variability to new version.

Thanks

PR #111, opened 2016-05-20
Rename Html:add() to Html:addHtml()

JanTvrdik:
https://twitter.com/jakubvrana/status/733655999618404352

### DIFF
--- a/src/Utils/Html.php
+++ b/src/Utils/Html.php
@@ -363,13 +363,35 @@ class Html implements \ArrayAccess, \Countable, \IteratorAggregate, IHtmlString
 
 
 	/**
+	 * @deprecated
+	 */
+	public function add($child)
+	{
+		trigger_error('Method Nette\Utils\Html::add() is deprecated, use addHtml() or addText() instead.', E_USER_DEPRECATED);
+		return $this->addHtml($child);
+	}
+
+
+	/**
 	 * Adds new element's child.
 	 * @param  Html|string Html node or raw HTML string
 	 * @return self
 	 */
-	public function add($child)
+	public function addHtml($child)
 	{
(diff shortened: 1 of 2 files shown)


src/Utils/Html.php
```diff
@@ -363,17 +363,39 @@ public function getText()
...
+	}
+
+
+	/**
 	 * Adds new element's child.
 	 * @param  Html|string Html node or raw HTML string
 	 * @return self
 	 */
-	public function add($child)
+	public function addHtml($child)
```
enumag: Maybe only allow Html instance?

fprochazka: Not sure that's a good idea. How would you provide custom HTML as string?

JanTvrdik: You would need to wrap it as an Html instance. But I think that the current solution is better.

JanTvrdik:
Other places which may lead to confusion:

* `$el->insert` – maybe not used by anyone directly
* `$el[]` – likely used by somebody, don't know how to solve except for deprecating the entire `ArrayAccess` implementation

src/Utils/Html.php
```diff
@@ -363,17 +363,39 @@ public function getText()
 
 
 	/**
+	 * @deprecated
+	 */
+	public function add($child)
```
milo: I like how is the add() short. What about to allow HTML instance and strings consider as a text? Not sure how to warn about change.

castamir:
@JanTvrdik 
`$el->insert()` might became protected or should at least be annotated as `@internal`
`$el[]` is a shortcut for adding child elements, isn't it? We can modify array access setter to use `addHtml()`... or we can deprecate this feature if noone actually use it :D

PR #120, opened 2016-09-03
FileSystem: fix mkdir race condition

pepakriz:
Directory can be created by concurrent process. Similar solution by Symfony in this issue https://github.com/symfony/symfony/pull/11726


dg:
Thanks

pepakriz:
Thanks too, but... your solution broke tests :(

dg:
Uh, I forgot to commit test.

PR #121, opened 2016-09-22
ObjectMixin: fix annotation parsing in traits

matej21:


dg:
What about to add support only for properties, because methods are depreacted?

PR #122, opened 2016-10-19
Image::place() converts percentage arguments to numeric value

PavelJurasek:
Fixes compatibility with PHP 7.1 as it caused `E_NOTICE: A non well formed numeric value encountered` in [Image.php:474](https://github.com/nette/utils/blob/master/src/Utils/Image.php#L474).


dg:
Please fix it here too https://github.com/PavelJurasek/utils/blob/efb19fe940d5f152174688ba25d4af0714fd737d/src/Utils/Image.php#L324-L336

PavelJurasek:
Done

PavelJurasek:
Is anything else needed?

dg:
abs() is missing

PavelJurasek:
Fixed

dg:
Thanks

src/Utils/Image.php
```diff
@@ -322,14 +322,14 @@ public function resize($width, $height, $flags = self::FIT)
 	public static function calculateSize($srcWidth, $srcHeight, $newWidth, $newHeight, $flags = self::FIT)
 	{
 		if (is_string($newWidth) && substr($newWidth, -1) === '%') {
-			$newWidth = (int) round($srcWidth / 100 * abs($newWidth));
+			$newWidth = (int) round($srcWidth / 100 * substr($newWidth, 0, -1));
```
PavelJurasek: https://github.com/nette/utils/commit/8683a0bb07d80359b0226dd53f0257b734a74b9d was merged. ```abs``` is present in that commit.

PR #123, opened 2016-11-05
Validators: added isListOfType()

### LINKED ISSUE #119: Validators: support for types int[] and string[]
How do you feel about supporting either specifically `int[]` and `string[]` (defined as _list_ where all values are of given type)?

Or we can generally support `anything[]`, but I don't think that's necessary.

t0his:
I would like to resolve #119.


JanTvrdik:
You either need to rename the method to `isIterableOfType($value, $type)` or check whether the value is a list (by calling `self::isList($list))`

TomasVotruba:
I would expect this method to support objects as well:

```php
Validators::isIterableOfType($items, 'Product');
```

For inspiration:

- in PHPUnit, there is `assertInternalType` for string/int/bools.
- in [beberlei/assert](https://github.com/beberlei/assert/blob/8276488eab645604a103df374c1e0b6374bb11b9/lib/Assert/Assertion.php#L56) is `allIsInstanceOf`

t0his:
I thought about it but decided the object should implement Travarsable if you want to iterate through its properties. As is in new php7.1 function is_iterable.

JanTvrdik:
@TomasVotruba I think that the example you've posted should work just fine with the current implementation.

dg:
What about `everyIs`?

TomasVotruba:
@JanTvrdik Ah, that is missing in tests, that suggest only scalar types are supported.

@t0his Could you please add test for simple object check? Sth like:

```php
test(function () {
	Assert::true(Validators::isIterableOfType([new Product], Product::class));
	Assert::false(Validators::isIterableOfType([new Product, new stdClass], Product::class));
});
```

t0his:
@TomasVotruba Sure i can add it, but shouldnt this be covered by is() tests?

@dg Ok i can rename it once i get home.

dg:
It was only a suggestion, hold on.

TomasVotruba:
@t0his Most part of the test is already coverage by is(), but that's not the goal
The test would be easier to understand, if both use cases - scalar and object - would be covered.

t0his:
@TomasVotruba Ok would you like to add some other use case?

JanTvrdik:
Maybe extract method `isIterable` + support `is($value, 'iterable')`?

t0his:
But the method will soon be redundant with php7.1.

JanTvrdik:
Yes, the method would use `is_iterable` and fallback to implementation in PHP in PHP < 7.1; probably best to make the method private to allow removal in future without breaking BC.

TomasVotruba:
@t0his Nope, I guess that's all I would expect this method to do. Thank you.

PR #129, opened 2017-01-11
Reflection::getParameterDefaultValue() fixed constants parsing in namespaced scope

xificurk:
- bug fix? yes
- new feature? no
- BC break? no
- doc PR: not needed

`Reflection::getParameterDefaultValue()` does not properly resolve global constants used as default parameter values in namespaced classes.


xificurk:
Related https://bugs.php.net/bug.php?id=73632

dg:
Thanks!

Issue #83, opened 2015-08-24
Json.decode().phpt fails under Debian's php5


Majkl578:
I am constantly getting failures in Json.decode().phpt:

```
Note: No php.ini is used.
PHP 5.6.12-1 | php5 -d 'extension=json.so' | 8 threads

-- FAILED: Nette\Utils\Json::decode() | tests/Utils/Json.decode().phpt
   Failed: array('' => 1) should be 
       ... array("\x00" => 1)
```

```
$ php5 -r 'echo JSON_C_VERSION, PHP_EOL;'
0.11.99
$ dpkg-query -W php5-json
php5-json   1.3.7-1
```

dg:
@Majkl578 Can you solve it somehow? I have no access to Debian.

JanTvrdik:
I don't think you can workaround this jsonc limitation (unsupported nul bytes in strings), so imho this is won't fix.

milo:
Agree with @JanTvrdik And since PHP 7.0 there is a jsond extension, afaik without this bug.

dg:
So I am closing it.

hrach:
This should be fixed in test to check for proper json ext type, shouldn't it?

dg:
Is there anyone with Debian, one could fix the test?

Majkl578:
I don't think it's worth fixing anymore, master is 7.0+ anyway...

milo:
@dg I'am. Testing right now v2.4 branch (master is 7.0+).

I'm getting 5 failing Image tests and one Json
```
-- FAILED: Nette\Utils\Json::decode() | tests/Utils/Json.decode().phpt
   Failed: Nette\Utils\JsonException was expected, but none was thrown

   in tests/Utils/Json.decode().phpt(30) Tester\Assert::exception()
```
which is different message than reported.

milo:
Skipping it...
```
-- FAILED: Nette\Utils\Json::decode() | tests/Utils/Json.decode().phpt
   Failed: Nette\Utils\JsonException with a message matching 'Syntax error'
                                    ... was expected but got 'unexpected end of data'

   in tests/Utils/Json.decode().phpt(35) Tester\Assert::exception()
```

skipping it...
```
-- FAILED: Nette\Utils\Json::decode() | tests/Utils/Json.decode().phpt
   Failed: Nette\Utils\JsonException with a message matching 'State mismatch (invalid or malformed JSON)'
                                    ... was expected but got 'unexpected character'

[...]

PR #131, opened 2017-01-18
tests: fix for JSON-C extension on Debian/Ubuntu

milo:
- bug fix? no, ref. #83
- new feature? no
- BC break? no

PR is for v2.4 branch.

This is only headless test fix. I didn't think about fixing Json class.


dg:
Great!

PR #133, opened 2017-02-01
Immutable DateTime [WIP]

dg:
- new feature? no
- BC break? yes

I am looking for way how move from mutable to immutable DateTime.

From my point of view, mutable DateTime is something you never want (of couse I may be wrong). And switch from mutable to immutable is big BC break. On the other hand, I found way how to discover bottlenecks ([like this one](https://github.com/nette/http/blob/master/src/Http/Helpers.php#L29-L31)) where mutable and immutable  behave differently and I can trigger error. So maybe the transition from one model to another would be possible.


Majkl578:
+1 for immutable version, but I don't think changing parent class is a good idea due to BC constranints, mostly because of typehints in userland and entirely different behavior... Rather properly deprecate DateTime in favor of new variant.

dg:
Deprecate for new variant seems better to me. The point is that a lot of people just will change class name, so functionality that traces bottlenecks should be useful in new class. 

I am not sure about name, because `Nette\Utils\DateTimeImmutable` sounds terrible.

TomasVotruba:
> I am not sure about name, because Nette\Utils\DateTimeImmutable sounds terrible.

Why? It sounds explicit and logic. 

- `DateTime` => `Nette\Utils\DateTime`
- `DateTimeImmutable` => `Nette\Utils\DateTimeImmutable`

dg:
`\DateTimeImmutable` sound terrible too :)

TomasVotruba:
Well, that's issue for another repository :)

michalhlavka:
What about something like that Nette\Utils\Immutable\DateTime, that also wrong, but something like this principle

Majkl578:
There is still #82 though - kill DateTime entirely. 😊

dg:
To kill DateTime is colossal BC break.

Majkl578:
Migration to different class is as well and helper methods would be kept in new helper class.
And honestly, what is the added value of extending (inheriting) PHP's DateTime(Immutable)? Adding some helper methods and making code less portable?

dg:
__toString and jsonEncode is added value.

dg:
Helper class for immutable DateTime can have name DateTimes, for cosistency with Arrays and Strings ;)

milo:
What about name `Timestamp`?

pavelkouril:
The helper class for the added functionality (working with `DateTimeInterface`) and just deprecating Nette `DateTime` is IMHO the best option.

The loss of `__toString` kinda sucks, but other aspects outweight the loss of it.

PR #135, opened 2017-02-13
Remove chars breaking valid JSON format

kastanekdavid:
Unsupported characters throw error:
... does not contain valid JSON
Parse error on line xxx:
.../nette.org/donate": "[1;37;42m Please c
----------------------^
Invalid string, it appears you forgot to terminate a string, or attempted to write a multiline string which is invalid

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
 		"ext-mbstring": "to use Strings::lower() etc...",
 		"ext-xml": "to use Strings::length() etc. when mbstring is not available",
 		"ext-gd": "to use Image",
-		"https://nette.org/donate": "\u001b[1;37;42m Please consider supporting Nette via a donation \u001b[0m"
+		"https://nette.org/donate": "Please consider supporting Nette via a donation"
 	},
 	"conflict": {
 		"nette/nette": "<2.2"


dg:
Where this error appears?

kastanekdavid:
During `composer update` with private Satis repository.

I use [satis](https://github.com/composer/satis) for 3rd library caching. Satis generates huge JSON file with dependencies of all composer packages used in my projects. While Composer parses the file using [JsonParser](https://github.com/Seldaek/jsonlint/blob/master/src/Seld/JsonLint/JsonParser.php), it throws exception on the line above. By definition composer tries to repeat this 3 times and then fails.

dg:
It is interesting, file seems fine, `composer validate --strict` tells its ok“…

kastanekdavid:
The JSON is valid. So I will issue a bug report for [https://github.com/Seldaek/jsonlint](JSONLint) project. Thanks.

Majkl578:
I don't see any error using the abovementioned library:
```
$ curl -Ls https://raw.githubusercontent.com/nette/utils/master/composer.json > json
$ composer require seld/jsonlint @dev
./composer.json has been created
Loading composer repositories with package information
Updating dependencies (including require-dev)
  - Installing seld/jsonlint (dev-master bbb5a42)
    Cloning bbb5a42bbdbf3f1e7d64b43443078d99e060ecf7 from cache

Writing lock file
Generating autoload files
$ vendor/bin/jsonlint json 
Valid JSON
```

kastanekdavid:
Running the above on PHP 5.3 might [clear this out](http://phpsadness.com/sad/20) :D

PR #136, opened 2017-03-02
Destroy resource if Image is destroyed [WIP]

bckp:
If you do not destroy image with imagedestroy, resource will be allocated for the rest of script life, this is problem in CLI scripts

- bug fix? yes
- new feature? no
- BC break? yes

This is mostly problem for CLI based scripts, that work with Nette\Utils\Images, if you save the image and unset variable, imagedestroy is not called and image stay in memory to the end of script life, it is fine if you load few images, but if you load more then few (10+) you can exhausted memory limit easily.

### DIFF
--- a/src/Utils/Image.php
+++ b/src/Utils/Image.php
@@ -219,7 +219,18 @@ class Image
 		imagesavealpha($image, TRUE);
 	}
 
+	
+	/**
+	 * Destroy image resource if any.
+	 */
+	public function __destruct()
+	{
+		if (is_resource($this->image)) {
+			imagedestroy($this->image);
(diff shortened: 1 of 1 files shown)


bckp:
Ummm, do not understand why tests fail :/

dg:
Because it is BC break.

bckp:
Yeah, I was testing master instead of my branche, this is why I was confused.

dg:
I don't understand why PHP doesn't release resource automatically? It should be reported as PHP bug.

dg:
closing due to human timeout

PR #137, opened 2017-03-07
Posibility to change indent character of rendering HTML attribute

aleswita:
- bug fix? no
- new feature? yes
- BC break? no

Possibility to set indent character in render method.


aleswita:
child bug fix + added test

dg:
Do you think that you will get a higher salary this way? https://stackoverflow.blog/2017/06/15/developers-use-spaces-make-money-use-tabs/

PR #139, opened 2017-05-09
Filesystem::write() accept NULL in $mode parameter

jkuchar:
- bug fix? yes
- new feature? no
- BC break? no (fixing unintended Nette 2.x BC break)

Code of function expects `$mode` to be `NULL`. If that happens it does not `chmod` it.

### DIFF
--- a/src/Utils/FileSystem.php
+++ b/src/Utils/FileSystem.php
@@ -21,11 +21,10 @@ final class FileSystem
 
 	/**
 	 * Creates a directory.
-	 * @param  int|NULL $mode
 	 * @return void
 	 * @throws Nette\IOException
 	 */
-	public static function createDir(string $dir, $mode = 0777)
+	public static function createDir(string $dir, int $mode = 0777)
 	{
 		if (!is_dir($dir) && !@mkdir($dir, $mode, TRUE) && !is_dir($dir)) { // @ - dir may already exist
 			throw new Nette\IOException("Unable to create directory '$dir'. " . error_get_last()['message']);
--- a/src/Utils/FileSystem.php
+++ b/src/Utils/FileSystem.php
@@ -131,14 +131,12 @@ final class FileSystem
 
 	/**
 	 * Writes a string to a file.
-	 * @param $mode int|NULL
+	 * @param int|NULL $mode File creation mode
 	 * @return void
 	 * @throws Nette\IOException
 	 */
 	public static function write(string $file, string $content, $mode = 0666)
 	{
-		assert($mode === NULL || is_int($mode));
-
 		static::createDir(dirname($file));
 		if (@file_put_contents($file, $content) === FALSE) { // @ is escalated to exception
 			throw new Nette\IOException("Unable to write file '$file'.");
--- a/src/Utils/FileSystem.php
+++ b/src/Utils/FileSystem.php
@@ -131,7 +131,7 @@ final class FileSystem
 
 	/**
 	 * Writes a string to a file.
-	 * @param int|NULL $mode File creation mode
+	 * @param int|NULL $mode
 	 * @return void
 	 * @throws Nette\IOException
 	 */
--- a/src/Utils/FileSystem.php
+++ b/src/Utils/FileSystem.php
@@ -21,10 +21,11 @@ final class FileSystem
 
 	/**
 	 * Creates a directory.
+	 * @param  int|NULL $mode
 	 * @return void
 	 * @throws Nette\IOException
 	 */
-	public static function createDir(string $dir, int $mode = 0777)
+	public static function createDir(string $dir, $mode = 0777)
 	{
 		if (!is_dir($dir) && !@mkdir($dir, $mode, TRUE) && !is_dir($dir)) { // @ - dir may already exist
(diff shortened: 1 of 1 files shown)


dg:
It breaks compatibility with PHP 7.0

dg:
Ping @jkuchar

src/Utils/FileSystem.php
```diff
@@ -131,11 +131,14 @@ public static function read(string $file): string
 
 	/**
 	 * Writes a string to a file.
+	 * @param $mode int|NULL
```
jkuchar: Not sure how to write php doc here as you usually do not use @param with $variableName.

src/Utils/FileSystem.php
```diff
@@ -131,11 +131,14 @@ public static function read(string $file): string
...
 	 * @return void
 	 * @throws Nette\IOException
 	 */
-	public static function write(string $file, string $content, int $mode = 0666)
+	public static function write(string $file, string $content, $mode = 0666)
 	{
+		assert($mode === NULL || is_int($mode));
+
```
jkuchar: I usually use assert when there is no possibility of declaring type using typehint (union types, ...) Do you like it or should it fail at line [144](https://github.com/nette/utils/pull/139/files#diff-f0409c8a1df5e9194764902390ef6f9bR144)

src/Utils/FileSystem.php
JanTvrdik: Most importantly you switched parameter name with parameter type, the annotation should be `@param int|NULL $mode`

src/Utils/FileSystem.php
JanTvrdik: Seems unnecessary to me.

jkuchar:
@JanTvrdik thanks for review, updated

src/Utils/FileSystem.php
```diff
@@ -131,10 +131,11 @@ public static function read(string $file): string
 
 	/**
 	 * Writes a string to a file.
+	 * @param int|NULL $mode
```
jkuchar: It does not look obvious what this parameter does... Is that creation mode or does it override current file mode?

Maybe add to docs something like:
- Alters file mode to given value (does not matter if file already or will be created exists)

**Shouldn't NULL be the default value?** Current default value is not secure at all. It allows everyone on the server to read and write to this file. And imagine situation where that server is shared... Why not to leave this value to `NULL` and keep responsibility of setting proper permissions on system admin that configured that server? In my case `chmod(0666)` basically breaks what was set by ACL and allows reading and writing by everyone.

jkuchar:
@JanTvrdik Should I squash this, have you finished review?

src/Utils/FileSystem.php
JanTvrdik: There should be two spaces after `@param` to align with `@return` and `@throws`.

jkuchar:
@JanTvrdik Added the same fix also for `createDir()` e5ecd3708428ec61ad8f32fb89135a27b2d6e0f0 Anything more that needs to be fixed?

JanTvrdik:
@jkuchar Hopefully not.

dg:
How differs `mkdir($dir, 0777)` from `mkdir($dir, NULL)`?

jkuchar:
@dg **you are right.**

I have performed the following test.

```php
mkdir("testFromPHPNULL", NULL);
mkdir("testFromPHP0777", 0777);
exec("mkdir -m 0777 testFromBash0777");
exec("mkdir testFromBashDefault");
```

Tried this obscure situation with ACLs:
```bash
$ getfacl .                                 
# file: .
# owner: jkuchar1a
# group: domain^users
user::rwx
group::r-x                      #effective:---
mask::---
other::---
default:user::rwx
default:group::r-x              #effective:---
default:mask::---
default:other::---
```

```bash
$ ll
drwx------+  2 jkuchar1a domain^users 4096 čen  9 19:40 testFromBashDefault/
drwx------+  2 jkuchar1a domain^users 4096 čen  9 19:40 testFromBash0777/
d---------+  2 jkuchar1a domain^users 4096 čen  9 19:40 testFromPHPNULL/  <--- WRONG
drwx------+  2 jkuchar1a domain^users 4096 čen  9 19:40 testFromPHP0777/
```

Result permissions on **testFromBashDefault** and **testFromPHP0777** are the same.

When no ACLs are used it behaves the same.

```bash
$ ll
drwxr-xr-x  2 jkuchar1a domain^users 4096 čen  9 19:46 ./
drwxr-xr-x 12 jkuchar1a domain^users 4096 čen  9 19:26 ../
```

```bash
drwxr-xr-x  2 jkuchar1a domain^users 4096 čen  9 19:50 testFromBashDefault/
drwxrwxrwx  2 jkuchar1a domain^users 4096 čen  9 19:50 testFromBash0777/ <--- WRONG
d---------  2 jkuchar1a domain^users 4096 čen  9 19:50 testFromPHPNULL/ <--- WRONG
drwxr-xr-x  2 jkuchar1a domain^users 4096 čen  9 19:50 testFromPHP0777/
```

dg:
Thanks

PR #142, opened 2017-06-19
Ability to insert new item into array without specifying key

CZechBoY:
- bug fix? no   
- new feature? yes
- BC break? yes

With this PR this is possible:
```php
$arr = [1,2,3];
Arrays::insertAfter($arr, 2, [4]);
```
instead of 
```php
$arr = [1,2,3];
Arrays::insertAfter($arr, 2, [3 => 4]);
```


CZechBoY:
As I see that array_merge reindexes the new array and there is only one alternative (+ operator), then I close this pr and make own functions.

PR #143, opened 2017-06-26
Mapper: supports count

CZechBoY:
- bug fix? no
- new feature? yes
- BC break? no

Now `count($mapper)` returns always `1` and no error is triggered (until 7.2.0).
https://3v4l.org/56k0a

Current workaround is to use the code from `count` method.

I can add iterator_count to CachingIterator too... Or if this behavior is intended then I can copy it to Mapper to support it this way.

### DIFF
--- a/src/Iterators/Mapper.php
+++ b/src/Iterators/Mapper.php
@@ -15,7 +15,7 @@ use Nette;
 /**
  * Applies the callback to the elements of the inner iterator.
  */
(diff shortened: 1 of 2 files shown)


dg:
The mapper's purpose is not to add the Counttable interface.

I know that in PHP <7.2 works `count()` badly, but since it affects all iterable objects, there is no need to fix this behavior in the Mapper.

PR #144, opened 2017-07-12
Paginator::getLength returns only int

matej21:


Majkl578:
property-read was not updated.

dg:
fixed

PR #147, opened 2017-08-20
Strings: added method for bytes counting

scheras:
- bug fix? no
- new feature? yes
- BC break? no
- doc PR: nette/docs#632

Adds new method useful for bytes counting.


JanTvrdik:
* `Strings` class is designed for UTF-8 encoding, using the same class for handling bytes is confusing
* mbstring.func_overload is [deprecated since PHP 7.2](https://wiki.php.net/rfc/deprecations_php_7_2)

dg:
This is probably useless because it's easier to use strlen().

Issue #149, opened 2017-10-06
Strings::normalize() should normalize into UTF-8 NFC


jkuchar:
- bug report? no
- feature request? yes
- version: 2.4

### Description
There are [more ways](https://stackoverflow.com/a/7934397) how to save UTF-8 strings. This means that when one want to compare two strings it can be not that simple and intuitive.

```php
$char_A_ring = "\xC3\x85"; // Å 'LATIN CAPITAL LETTER A WITH RING ABOVE' (U+00C5)
$char_combining_ring_above = "A\xCC\x8A";  // Å 'COMBINING RING ABOVE' (U+030A)

assert(\Nette\Utils\Strings::compare($char_A_ring, $char_combining_ring_above)); //error
assert(\Nette\Utils\Strings::normalize($char_A_ring) === \Nette\Utils\Strings::normalize($char_combining_ring_above)); //error

assert($char_A_ring === $char_combining_ring_above);  //error
```
I would expect that when I do this after Strings::normalize() this would succeed, but it does not. I would expect to normalize everything into NFC or NFD (does not matter that much). Currently there is no guarantee that two same strings will be understood as the same.

jkuchar:
I would be happy to provide patch as this is very simple to implement with [intl extension](http://php.net/manual/en/normalizer.normalize.php).

dg:
So can you add `normalizeUtf8()` function and use it in `compare()`?

dg:
@jkuchar can this be closed now?

jkuchar:
Thanks for nice collaboration! :-)

PR #150, opened 2017-10-14
Strings: normalize strings into UTF-8 NFC

jkuchar:
- bug fix? no   
- new feature? yes
- BC break? yes

implementation of #149 

Before merge:
- [ ] decide what to do with compare/normalize [inconsistency](https://github.com/nette/utils/pull/150#discussion_r144794301)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,8 +15,7 @@
 		}
 	],
 	"require": {
-		"php": ">=7.0",
-		"ext-intl": "*"
+		"php": ">=7.0"
 	},
 	"require-dev": {
 		"nette/tester": "~2.0",
@@ -25,7 +24,7 @@
 	"suggest": {
 		"ext-iconv": "to use Strings::webalize() and toAscii()",
 		"ext-json": "to use Nette\\Utils\\Json",
-		"ext-intl": "for script transliteration in Strings::webalize() and toAscii()",
+		"ext-intl": "to use Strings::webalize(), toAscii(), normalize() and compare()",
 		"ext-mbstring": "to use Strings::lower() etc...",
 		"ext-xml": "to use Strings::length() etc. when mbstring is not available",
 		"ext-gd": "to use Image"
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,8 @@
 		}
 	],
 	"require": {
-		"php": ">=7.0"
+		"php": ">=7.0",
+		"ext-intl": "*"
 	},
 	"require-dev": {
 		"nette/tester": "~2.0",
--- a/src/Utils/Strings.php
+++ b/src/Utils/Strings.php
@@ -99,10 +99,13 @@ class Strings
 
 
 	/**
-	 * Removes special controls characters and normalizes line endings and spaces in UTF-8 string.
+	 * Removes special controls characters and normalizes line endings and converts to NFC form and spaces in UTF-8 string.
 	 */
 	public static function normalize(string $s): string
 	{
+		// normalize string into utf8 NFC form
+		$s = \Normalizer::normalize($s, \Normalizer::FORM_C);
+
 		$s = self::normalizeNewLines($s);
 
 		// remove control characters; leave \t + \n
@@ -265,6 +268,9 @@ class Strings
 	 */
 	public static function compare(string $left, string $right, int $len = null): bool
 	{
+		$left = \Normalizer::normalize($left, \Normalizer::FORM_D); // form NFD is faster
+		$right = \Normalizer::normalize($right, \Normalizer::FORM_D); // form NFD is faster
+
 		if ($len < 0) {
 			$left = self::substring($left, $len, -$len);
 			$right = self::substring($right, $len, -$len);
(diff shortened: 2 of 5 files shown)


composer.json
```diff
@@ -15,7 +15,8 @@
 		}
 	],
 	"require": {
-		"php": ">=7.0"
+		"php": ">=7.0",
+		"ext-intl": "*"
```
JanTvrdik: should probably remain only as suggested

src/Utils/Strings.php
```diff
@@ -99,10 +99,13 @@ public static function substring(string $s, int $start, int $length = null): str
 
 
 	/**
-	 * Removes special controls characters and normalizes line endings and spaces in UTF-8 string.
+	 * Removes special controls characters and normalizes line endings and converts to NFC form and spaces in UTF-8 string.
```
JanTvrdik: The sentence now makes no sense. How about sth like

Removes special controls characters and normalizes line endings, spaces and normal form to NFC in UTF-8 string

src/Utils/Strings.php
```diff
@@ -99,10 +99,13 @@ public static function substring(string $s, int $start, int $length = null): str
...
 	 */
 	public static function normalize(string $s): string
 	{
+		// normalize string into utf8 NFC form
```
JanTvrdik: * „UTF-8“ instead of „utf8“, or maybe even better to omit altogether
* the word „string“ is unnecessary

How about sth like „convert to NFC normal form “?

jkuchar:
@JanTvrdik Thanks for comments 😃

composer.json
jkuchar: What should `Strings` class do when there is no intl extension?

src/Utils/Strings.php
jkuchar: 👍

src/Utils/Strings.php
jkuchar: 👍 has been writing it in a car

composer.json
jkuchar: @hrach could you please elaborate on it more then just thumbs down? I think I know some of your arguments, but want to be sure what is on your mind.

src/Utils/Strings.php
jkuchar: fixed in b6dd526

src/Utils/Strings.php
jkuchar: fixed in ec16e40

composer.json
hrach: Well, I consider the Strings class as the main functionality of the nette/utils package and therefore it's dependencies should be required. Nothing more, nothing less.

JanTvrdik: @hrach That's a valid opinion, but it's outside of scope of this PR. If you want to move currently suggested stuff to `require` you need another PR and do it consistently for all stuff in `suggest`.

jkuchar: That makes sense, as that was my next question - what to do with the other dependencies.

jkuchar: fixed in bb3a69cac366baa5ec7a5875099e12a12b48f9e2

jkuchar:
@JanTvrdik @hrach thanks for nice argument-based discussion!

I have pushed fixes into this pull request, are there any more comments on this?

src/Utils/Strings.php
```diff
@@ -265,6 +268,9 @@ public static function capitalize(string $s): string
 	 */
 	public static function compare(string $left, string $right, int $len = null): bool
 	{
+		$left = \Normalizer::normalize($left, \Normalizer::FORM_D); // form NFD is faster
+		$right = \Normalizer::normalize($right, \Normalizer::FORM_D); // form NFD is faster
```
jkuchar: Does it make sense to call Normalizer directly? Or would we prefer to call self::normalize(). This will also normalizer line ending and some other stuff. I'm not sure on which level should be normalization done.

Current behaviour (proposed by @dg) causes inconsistencies as it is possible to get into situation where:

```php
Strings::compare($a, $b); // FALSE
Strings::normalize($a) === Strings::normalize($b); // TRUE
```

Is this expected behaviour? It seems confusing to me.

composer.json
jkuchar: openned #151

src/Utils/Strings.php
dg: Is it possible to call it after substring() ?

JanTvrdik: No. Substring operates on code units, normalize may change length in code units (imho).

jkuchar:
The last thing to sort out is the following inconsistency. Current behaviour (proposed by @dg) causes inconsistencies as it is possible to get into situation where:

```php
Strings::compare($a, $b); // FALSE
Strings::normalize($a) === Strings::normalize($b); // TRUE
```

It seems as confusing behaviour to me. What are your opinions on this?

JanTvrdik:
> Current behaviour (proposed by @dg) causes inconsistencies

This seems OK to me. It would be quite confusing if `Strings::compare("\r\n", "\t   \x00\n")` returned true. The `compare` function is just case-insensitive variant of `===`.

jkuchar:
👍 Ok, it makes sense. So, now it looks like **ready to merge**,

dg:
Thank you

PR #155, opened 2017-11-21
FileSystem::rename() fix renaming file/directory if only case changes

Ciki:
- bug fix? yes
- new feature? no
- BC break? no

When trying to rename file/directory itself in order to change only case of filename (eg. `test.php` to `Test.php`), the process fails as the method first [deletes](https://github.com/nette/utils/blob/master/src/Utils/FileSystem.php#L110) the desired file(name) prior to renaming it .. thus it deletes the original file/dir that was about to be renamed.
This change prevents that case.

### DIFF
--- a/src/Utils/FileSystem.php
+++ b/src/Utils/FileSystem.php
@@ -107,7 +107,7 @@ final class FileSystem
 
 		} else {
 			static::createDir(dirname($newName));
-			if (Strings::lower(realpath($name)) !== Strings::lower(realpath($newName))) {
+			if (realpath($name) !== realpath($newName)) {
 				static::delete($newName);
 			}
 			if (!@rename($name, $newName)) { // @ is escalated to exception
--- a/src/Utils/FileSystem.php
+++ b/src/Utils/FileSystem.php
@@ -107,7 +107,7 @@ final class FileSystem
 
 		} else {
 			static::createDir(dirname($newName));
-			if (Strings::lower($name) !== Strings::lower($newName)) {
+			if (Strings::lower(realpath($name)) !== Strings::lower(realpath($newName))) {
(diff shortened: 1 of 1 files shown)


dg:
Good point. However, checking whether the files are the same should be done differently (I don't know how). For compatibility between NTFS and Linux.

Ciki:
@dg you think `realpath` could do the work properly? I added new commit prior to seeing your comment

dg:
Maybe compare `stat`?

Ciki:
I think the `realpath` should do the work properly but I am no expert on this.  Any use case where it might break? I can add the `stat` comparison but not sure when it can fail and return `false` thus skipping the removal of dest file (eg. when FALSE is returned for both different source and dest file)

dg:
Realpath should do the work, I didn’t test it. It this case there shouldn’t be lower().

Ciki:
@dg updated

dg:
Did you test it? I am not sure that realpath really works this way.

Ciki:
@dg I did.. on OS preventing same filenames differing in case only, like Windows, realpath correctly transforms eg `Test.php` and `teSt.php` to `test.php` if `test.php` exists. On the other hand on Linux, where filenames can differ in case, it returns the original and correct names. As far as I can tell and test, it works correctly

dg:
Great. But it seems that it breaks test…

Ciki:
@dg should I remove the failing test altogether? As the failing `rename` method cannot be actually tested IMO - according to [php.net spec](http://php.net/manual/en/function.rename.php) `If renaming a directory and newname exists, this function will emit a warning.` but this case never occurs

dg:
Thanks, merged

PR #156, opened 2017-11-24
Reflection::getParameterDefaultValue: fix constant visibility handling

### LINKED ISSUE #154: Nette\Utils\Reflection::getParameterDefaultValue does not work with private/protected constants
- bug report? yes
- feature request? no

```
Error: Cannot access protected const Foo::BAR
```

is there any reason why `getDefaultValue` cannot be used directly?

matej21:
- bug fix? yes
- new feature? no
- BC breakno
- fixes #154


dg:
Thanks!

PR #157, opened 2017-11-30
add PHPStan and fix found issues

JanTvrdik:


f3l1x:
Good job. One thing bother me, it means that only nette/utils has defined propel phpDoc and other packages have not. Do you plan at it to the others, too?

JanTvrdik:
I don't see how that is relevant. It it forbidden to fix bug in one package without fixing bugs in all packages?

f3l1x:
Nope, don't get me wrong I like the change.

```php
	/**
	 * Replaces or appends a item.
	 * @param  int|null
	 * @return void
	 * @throws Nette\OutOfRangeException
	 */
	public function offsetSet($index, $value)
```

=>

```php
	/**
	 * Replaces or appends a item.
	 * @param  int|null $index
	 * @return void
	 * @throws Nette\OutOfRangeException
	 */
	public function offsetSet($index, $value)
```

But, we should do it for all packages. To keep it same. 

I don't want to add phpdoc to one method and not to others, right?

dg:
Great work. I merged most of issues 32fd4778fb2a6f3654577e4c853d76f8c9f1e538

I have not added PhpStan yet, I must first become familiar with it.

Also I did not add the parameter $names in phpDoc because it is not used in another places.

JanTvrdik:
> Also I did not add the parameter $names in phpDoc because it is not used in another places.

What places? You mean other `nette/*` packages?

dg:
Yes.

JanTvrdik:
@dg Well, we can't add this to all packages at once. One package will always need to be the first. Also we've already had a related [discussion few years ago](https://forum.nette.org/en/21174-rfc-writing-parameter-names-in-param-annotations-accepted).

dg:
Yes, RFC was accepted.

dg:
@JanTvrdik how can be solved this error? https://travis-ci.org/nette/utils/jobs/361331623

Majkl578:
Needs to be added to ignored errors: https://github.com/nette/utils/blob/master/phpstan.neon#L2

JanTvrdik:
@dg Here are your options:

* fix PHP configuration on Travis, so that the `imagewebp()` function actually exists
* provide dummy (stub) definition of `imagewebp()` in PHPStan bootstrap file
* ignore the error

dg:
How can I fix PHP configuration on Travis?

Majkl578:
I don't think you can _fix_ it, that's why I didn't suggest it. PHP must be compiled with WEBP support, but it obviously isn't on Travis. :/
http://php.net/manual/en/image.installation.php

PR #159, opened 2018-01-02
Fix typo in exception message

bobey:
- bug fix? no
- new feature? no
- BC break? no

Fix a typo in an exception message thrown when character list contains less than two chars.


dg:
Please fix test too.

dg:
Thanks

PR #161, opened 2018-01-17
Added removeAttributes method - removing multiple attributes

Machy8:
- bug fix? no 
- new feature? yes

**Description**
- Added method removeAttributes for removing multiple attributes
- Why: for example when converting html element to the Amp html element, sometimes is necessary to remove multiple (by Amp disabled) attributes like frameborder, allowfullscreen or scrolling
- There is already a method for setting multiple attributes but non for removing multiple attributes so it means to create loops in order to remove them


dg:
Thanks

PR #165, opened 2018-02-26
FileSystem: added move()

hermajan:
- bug fix? no
- new feature? yes
- BC break? no
- doc PR: nette/docs#680

I added FileSystem::move() method for moving files.

### DIFF
--- a/src/Utils/FileSystem.php
+++ b/src/Utils/FileSystem.php
@@ -89,6 +89,17 @@ final class FileSystem
 
 
 	/**
+	 * Moves a file or directory.
+	 * @throws Nette\IOException
+	 */
+	public static function move(string $source, string $dest, bool $overwrite = true): void
+	{
+		static::copy($source, $dest, $overwrite);
+		static::delete($source);
+	}
+
+
+	/**
 	 * Renames a file or directory.
 	 * @throws Nette\IOException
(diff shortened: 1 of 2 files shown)


src/Utils/FileSystem.php
```diff
@@ -88,6 +88,17 @@ public static function delete(string $path): void
 	}
 
 
+	/**
+	 * Moves a file or directory.
+	 * @throws Nette\IOException
+	 */
+	public static function move(string $source, string $dest, bool $overwrite = true): void
+	{
+		static::copy($source, $dest, $overwrite);
```
dg: Vhodnější je použít rename(), protože copy() je mnohem pomalejší.

nechutny: Better is try rename and then if it fail (eg. two filesystems) try copy.

dg:
What is a difference between FileSystem::move() and FileSystem::rename()?

dg:
ping @hermajan

hermajan:
I did this function because from the source code and documentation I didn't know that FileSystem::rename() function moves files. Because it looks like the function of this pull request is unnecessary, I would suggest it might be worth mentioning explicitly in the documentation that FileSystem::rename() can be used to move files or make a FileSystem::move() alias of rename() for moving of files.

dg:
Ok, so I close it.

PR #167, opened 2018-02-28
Used native PHP function for image types

holantomas:
Used `image_type_to_extension` and `image_type_to_mime_type`

- bug fix? no
- new feature? no
- BC break? no

I think it's better to use native functions instead of work with arrays.
http://php.net/manual/en/function.image-type-to-mime-type.php
http://php.net/manual/en/function.image-type-to-extension.php


dg:
image_type_to_extension() returns jpeg or jpg? Will it work with imagefromjpeg()?

holantomas:
Yes and yes. Didn't find a way to get only jpg from this function.

PR #170, opened 2018-04-05
Image: throw exception when $file and $type is null

vlastavesely:
Fixes problem with calling 'pathinfo()' with null filename.

- bug fix? yes
- BC break? no

### DIFF
--- a/src/Utils/Image.php
+++ b/src/Utils/Image.php
@@ -483,6 +483,10 @@ class Image
 	 */
 	public function save(string $file = null, int $quality = null, int $type = null): void
 	{
+		if ($type === null && $file === null) {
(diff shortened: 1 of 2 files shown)


vlastavesely:
`Image::send()` calls the `save()` function with `$file` being set to `NULL` and you can never be sure whether someone does not use the same way in their application. I think that it would be a BC. My solution just converts confusing typehint error to more understandable exception.

vlastavesely:
@dg  What does it solve? `Image::save()` still can be called with no arguments and thus emit `TypeError (pathinfo() expects parameter 1 to be string, null given)`.

dg:
Fixed

PR #171, opened 2018-04-24
Removed the \t char for 4 spaces.

tiagosampaio:
- bug fix? no   
- new feature? no
- BC break? no

Replaced the \t character for 4 spaces by default. It's not a good practice to use tabs on indentation.

### DIFF
--- a/src/Utils/Strings.php
+++ b/src/Utils/Strings.php
@@ -206,13 +206,26 @@ class Strings
 		}
 		return $s;
 	}
-
-
-	/**
-	 * Indents UTF-8 string from the left.
-	 */
-	public static function indent(string $s, int $level = 1, string $chars = "    "): string
+    
+    
+    /**
+     * Indents UTF-8 string from the left.
+     *
+     * @param string $s
+     * @param int    $level
+     * @param string $chars
+     * @param int    $charMultiplier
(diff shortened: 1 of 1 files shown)


dg:
I prefer 3 spaces. Why four?

milo:
Sorry, but three or four spaces is nonsence. The best practice is 7.

dg:
It is true that 7 is my lucky number… I'm confused, because I really prefer three spaces.

Solution may be third parameter, where everybody can pass the string with how many spaces he wants. Default value can be tab, because it is always one. @tiagosampaio, what do you think?

tiagosampaio:
@dg I think it's a great idea to create a new parameter for that reason. It solves the problem (and preferences) of us all :D

tiagosampaio:
@milo Really? 7 spaces for each indentation? You've got to implement and follow the object calisthenics principles straightforwardly for every piece of code you create or you have to change the code standard to use 250 columns... I'm just kidding bro. :D

dg:
@tiagosampaio maybe this will help you https://github.com/nette/php-generator#tabs-versus-spaces

tiagosampaio:
Thanks @dg! I think it will help me.

Issue #168, opened 2018-03-04
Using Validators in native assert


zeleznypa:
It can be useful, when the [Validators::assert](https://github.com/nette/utils/blob/e31c8ff28e07f499c41ee7117c4ba4c5e0a5c761/src/Utils/Validators.php#L75) will return `TRUE` when the check is success.

After that it can be used in native [`assert`](http://php.net/manual/en/function.assert.php) function like in the following example:

```php
/**
 * Something getter
 *
 * @return ISomething
 * @throws AssertionException
 */
public function getSomething(): ISomething
{
    assert(Validators::assert($this->something, ISomething::class, 'something before get'));
    return $this->something;
}
```

In this way, the code can be more optimised in production by [`zend.assertions = -1`](http://php.net/manual/en/ini.core.php#ini.zend.assertions)

dg:
Can you send PR?

JanTvrdik:
@zeleznypa You can already use `assert(Validators::is(...));`

zeleznypa:
@JanTvrdik You can, but it make different behaviour. 

When you have code like:
```php
 assert(Validators::assert($this->something, ISomething::class, 'something before get'));
```
You will get very nice exception with all needed information (*what you expect, what you get, ...*) [See](https://3v4l.org/ThkQb)

But when you use following:
```php
 assert(Validators::is($this->something, ISomething::class));
```
you will get just a part of that informations. [See](https://3v4l.org/hpuLq)

zeleznypa:
In the meaning of the static analysis, it is wrong to negative compare of the `void` return type.
https://phpstan.org/r/0c4c2fbc5c999e6ff076a8c03643b6a4

PR #173, opened 2018-06-09
Assertable asserts (#168)

zeleznypa:
- new feature: #168
- BC break: no
- doc PR: no


mstefan21:
Today I implemented same functionality in app what now working on it, because we need more information for testers, but don`t need it in production.

PR #174, opened 2018-07-09
Html::setText: encode quotes too

CZechBoY:
- bug fix
- BC break? yes

Hi,
first of all one question: Why was encoding of quotes skipped?

### DIFF
--- a/src/Utils/Html.php
+++ b/src/Utils/Html.php
@@ -352,7 +352,7 @@ class Html implements \ArrayAccess, \Countable, \IteratorAggregate, IHtmlString
 	public function addText($text)
 	{
 		if (!$text instanceof IHtmlString) {
-			$text = htmlspecialchars((string) $text, ENT_NOQUOTES, 'UTF-8');
+			$text = htmlspecialchars((string) $text, ENT_QUOTES, 'UTF-8');
 		}
 		return $this->insert(null, $text);
 	}
--- a/src/Utils/Html.php
+++ b/src/Utils/Html.php
@@ -317,7 +317,7 @@ class Html implements \ArrayAccess, \Countable, \IteratorAggregate, IHtmlString
 	final public function setText($text)
 	{
 		if (!$text instanceof IHtmlString) {
-			$text = htmlspecialchars((string) $text, ENT_NOQUOTES, 'UTF-8');
+			$text = htmlspecialchars((string) $text, ENT_QUOTES, 'UTF-8');
 		}
 		$this->children = [(string) $text];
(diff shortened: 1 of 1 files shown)


dg:
Because it is not needed in HTML text.

CZechBoY:
But it is needed when it is used in html attributes. Or is there any other way to create html parameter escaped text?

dg:
See documentation how to use Nette\Utils\Html

dg:
See documentation how to use Nette\Utils\Html

CZechBoY:
There is no mention to not use in html attributes, and there is no mention to use only in html body.

dg:
https://doc.nette.org/cs/2.4/html-elements#toc-obsah-elementu

CZechBoY:
So you mean I should create Html element with attribute containing another Html instance?
In my case I have Latte template which has element with attribute containing problematic Html instance.

dg:
I don't understand what you mean. setText() and setHtml() have nothing to do with attributes.

dg:
I'm not here to guess what Adam meant.

JanTvrdik:
> I'm not here to guess what Adam meant.

Is a viable approach for normal issues. It's not viable approach for security-related issues.

dg:
See https://github.com/nette/utils/blob/master/.github/ISSUE_TEMPLATE/Bug_report.md

CZechBoY:
Sorry, I did not read those issue templates, possibly because I craeted pull request instead of issue.

Ok here is example with working code.
```php
https://github.com/CZechBoY/html-parameter-xss/blob/master/test.php
```

dg:
Html is not supposed to be used this way.

JanTvrdik:
@CZechBoY You previously mentioned Latte, but the test does not use Latte. Can you post an example with Latte where you've encountered this problem?

CZechBoY:
@dg Then special chars will be double escaped, quotes only once.
@JanTvrdik I realized that it affects only version 2.3. Version 2.4 has context for html attributes.

JanTvrdik:
> Nowhere is mentioned that this is a security issue.

Should in the future sentence such as „escaping is a concept directly related to security“ or „wrong escaping is an indication of possible security issue“ be included? The user (in general) may very well not understand escaping well enough to know what is and what is not a security issue. Please be more careful in the future when closing security-related issues / pull requests. This issue just happened to be fixed in Latte 2.4, the next one may not be.

dg:
Be so kind and do not teach me how to deal with issues.

PR #180, opened 2018-12-07
Arrays: added wrap() method

milo:
- new feature
- BC break? no
- doc PR: will

I'm using such conctruction quite often, maybe I'm not the only one.

### DIFF
--- a/src/Utils/Arrays.php
+++ b/src/Utils/Arrays.php
@@ -343,4 +343,19 @@ class Arrays
 	{
 		return key([$value => null]);
 	}
+
+
+	/**
+	 * Returns copy of the $array where every item is converted to string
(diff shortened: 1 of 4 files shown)


milo:
Maybe implement it by
```php
count($arr) ? ($prefix . implode($suffix . $glue . $prefix, $arr) . $suffix) : '';
```
but crucial is an API. No need for array count condition and prefix/suffix on single place.

dg:
What about method `wrap($arr, $prefix, $suffix)`? And than combine `implode` with `wrap`?

milo:
Sure, `wrap()` is better approach. Updated.

Not sure about type casting everything to string.

dg:
Great

milo:
Thank you!

PR #182, opened 2019-01-23
try to fix "Strings::truncate" for invalid UTF-8

voku:
-> this should fix errors in "rectorphp/rector" and "phpstan/phpstan"
-> see https://github.com/rectorphp/rector/issues/921#event-2087392895

PS: the current code throws a error -> ```Nette\Utils\RegexpException: Malformed UTF-8 data (pattern: #^.{1,6}(?=[\s\x00-/:-@\[-`{-~])#us)``` (https://travis-ci.org/nette/utils/jobs/483287132) ... maybe we could also catch the error instead of refactoring the code?


JanTvrdik:
Imho the error is expected, `Strings::truncate` expects valid utf-8 input. If you're unsure where the input is valid utf-8, you can always use `Strings::truncate(Strings::fixEncoding($s), ...)`. If your data are not utf-8 (such is the case in phpstan), you should not use `Strings::truncate` at all.

voku:
Okay, thank you for the quick answer. :+1:

PR #184, opened 2019-03-06
Adds the plus character as a valid option in numeric strings

Amunak:
- bug fix? no
- new feature? yes
- BC break? no
- doc PR: n/a

For whatever reason numeric strings with explicitly positive values (i.e. `+1` instead of just `1`) have not been considered as valid numerics even though they are perfectly valid (in the sense that even forcibly changing their type to the respective numeric type works as expected).

This PR fixes that issue, adding consistency with PHP and the way numbers can be written to the Validator class.

I have marked this as "not bc breaking" because (1) this use case has not been covered by tests, so it doesn't change any and (2) nobody should accept `'-1'` as a valid value and expect `'+1'` to fail.

Please merge this with squash; I accidentally made one too many commits.

### DIFF
--- a/src/Utils/Validators.php
+++ b/src/Utils/Validators.php
@@ -178,7 +178,7 @@ class Validators
 	 */
 	public static function isNumericInt($value): bool
 	{
-		return is_int($value) || is_string($value) && preg_match('#^-?[0-9]+\z#', $value);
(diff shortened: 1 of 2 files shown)


Amunak:
It would appear that your Travis test image is broken.

dg:
Thanks

PR #188, opened 2019-03-29
More verbose Json::decodeToObject and Json::decodeToArray methods

finwe:
- new feature
- No BC break
- doc PR: https://github.com/nette/docs/pull/760

Introduces Json::decodeToObject and Json::decodeToArray methods for clearer code intent and error avoidance.

### DIFF
--- a/src/Utils/Json.php
+++ b/src/Utils/Json.php
@@ -54,4 +54,24 @@ final class Json
 		}
 		return $value;
 	}
+
+
+	/**
+	 * Decodes a JSON string to a stdClass
+	 * @return \stdClass
+	 */
+	public static function decodeToObject(string $json)
+	{
+		return (object) self::decode($json);
+	}
+
+
+	/**
+	 * Decodes a JSON string to an array
+	 * @return mixed[]
+	 */
+	public static function decodeToArray(string $json)
+	{
+		return self::decode($json, self::FORCE_ARRAY);
+	}
 }
--- a/tests/Utils/Json.decodeToArray().phpt
+++ b/tests/Utils/Json.decodeToArray().phpt
@@ -1,0 +1,16 @@
+<?php
+
+/**
+ * Test: Nette\Utils\Json::decode()
+ */
+
+declare(strict_types=1);
+
+use Nette\Utils\Json;
+use Tester\Assert;
+
+
+require __DIR__ . '/../bootstrap.php';
+
+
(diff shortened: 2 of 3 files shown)


src/Utils/Json.php
```diff
@@ -54,4 +54,23 @@ public static function decode(string $json, int $flags = 0)
...
 		return $value;
 	}
+
+	/**
+	 * Decodes a JSON string to a stdClass
+	 * @return \stdClass
+	 */
+	public static function decodeToObject(string $json)
+	{
+		return self::decode($json);
```
mabar: Imho `["foo","bar"]` would fail here. Typecasting is needed.

finwe: would fail where? $json is type hinted as a string as it is in the main decode method.

mabar: Output of json_decode depends on json given in input. You could enforce array with FORCE_ARRAY, but there is no option FORCE_OBJECT

finwe: json_decode returns stdClass with no second parameter given (or FALSE) or an associative array with second parameter set to true. Or am I still missing something?

finwe: The behaviour of Json::decode is the same (just since 3.0 the boolean second parameter changed to integer represented with Json::FORCE_ARRAY)

mabar: http://sandbox.onlinephpfunctions.com/code/e22ec5b9f52f2f6f7fb4104efcad5ba3247a9e48

finwe: Oh, I get it now. Non-associative arrays.

dg:
I think that this behaviour not what user expects:

```php
Assert::equal((object) ['a', 'b'], Json::decodeToObject('["a","b"]'));
```

finwe:
What if I ditch the decodeToObject method altogether and keep only the decodeToArray?

src/Utils/Json.php
```diff
@@ -54,4 +54,24 @@ public static function decode(string $json, int $flags = 0)
...
+	 */
+	public static function decodeToObject(string $json)
+	{
+		return (object) self::decode($json);
+	}
+
+
+	/**
+	 * Decodes a JSON string to an array
+	 * @return mixed[]
```
JanTvrdik: This is incorrect, the method may return e.g. string.

finwe:
Nah, this has more holes than pros.

PR #192, opened 2019-06-19
Fix incorrect time zone when using DateTime() constructor

temistokles:
- bug fix
- BC break? no

DateTime fixes in Nette do not protect the __constructor against using incorrect timezone. This pull request fixes that.

### DIFF
--- a/src/Utils/DateTime.php
+++ b/src/Utils/DateTime.php
@@ -33,6 +33,12 @@ class DateTime extends \DateTime
 	/** average year in seconds */
 	const YEAR = 31557600;
 
+	public function __construct($time = "now", $timezone = NULL) {
+		if ($timezone === NULL) {
+			$timezone = new \DateTimeZone(date_default_timezone_get());
+		}
+		parent::__construct($time, $timezone);
+	}
 
 	/**
 	 * DateTime object factory.


dg:
Can you explain exactly what it does?

temistokles:
It tries to overcome:
https://bugs.php.net/bug.php?id=74173
https://bugs.php.net/bug.php?id=67634

Basically as soon as you use ->getTimestamp() or similar method working with timestamps (that are in UTC), the global timezone gets reset to Europe/London (which itself is not the same as UTC, which is only more funny).

At least this seems to be the issue with our PHP 7.1 installation. The issue is not consistent though and does not seem to happen all the time (we use CET time, therefore we experience the issue as two DateTimes created at around the same time as being offset by 1 hour).

Unfortunatelly from further testing of this fix does, it does not seem to be sufficient. The only thing that really helped was to force explicit timezone in __construct, from, setTimestamp, createFromFormat methods.
You cannot even rely on date_default_timezone_get() returning the correct timezone, as once the bug occurs, date_default_timezone_get() returns incorrect timezone as well.

In our case it was enough to force one specific timezone (the website only has to support single timezone), but this is not a general solution. But it should be enough to define the timezone in config and force it this way.

dg:
> the global timezone gets reset to Europe/London

Do you mean default time zone in object? 

I don't understand much, but when it comes to the method getTimestamp(), maybe the solution is to override it to `return (clone $this)->getTimestamp()`.

dg:
closing due to human timeout

PR #193, opened 2019-06-25
Strings: Optimalize strlen() for better performance.

janbarasek:
- new feature
- BC break? no

Strings performance improvements when working with string length.


dg:
Thanks!

PR #194, opened 2019-06-25
Validators: Better performance and codestyle

janbarasek:
Operations priority might differ from what you expect and replace substr to better strpos.

- new feature
- BC break? no


dg:
strpos() searches the entire string, so it is definitely worse than substr()

PR #195, opened 2019-06-25
Add __toString() return statement + __sleep() return type.

janbarasek:
- new feature
- BC break? no

Changed:

- `__sleep` must return array. Documentation: https://www.php.net/manual/en/language.oop5.magic.php#object.sleep
- `__toString` missing return statement.


dg:
Thanks!

PR #196, opened 2019-06-25
Html: Use PHP 7.1 features in code style.

janbarasek:
- new feature
- BC break? no

Lot's of changes. Short list:

- `setName` '$isEmpty ?? isset(static::$emptyElements[$name])' can be used instead (reduces cognitive load) and have better performance.
- `insert::$index` Optional parameter before required.


dg:
Thanks

PR #197, opened 2019-06-25
FileSystem: copy: Add binary safe parameter

janbarasek:
- new feature
- BC break? yes

`copy`: The mode is not binary-safe ('b' is missing).


dg:
Thanks

PR #198, opened 2019-06-25
Arrays: Better performance

janbarasek:
- new feature
- BC break? no

Changed:

- `associate`: Avoid callables in loop conditionals for better performance.
- `associate`: Condition `$parts === ['->']` execution costs less than the previous one
- `normalize`: Duplicated method calls should be moved to a local variable for better performance.


dg:
strlen(), count(), is_*() are not a real functions, it is fast instruction. Compare this:

- https://3v4l.org/vqsqY/vld#output
- https://3v4l.org/ZE4na/vld#output

So, some of these optimizations actually improve performance negligibly, or may even make it worse (it takes a time to create a variable).

Of course, these "functions" must be called in global namespace, so I optimized code few years ago via 237a9096f8280b6648cd8343ba70ac1c055d81c8, but now I see that `count` is missing.

PR #200, opened 2019-08-28
Added missing reflection types

mabar:
- bug fix / new feature? both, I guess
- BC break? no

### DIFF
--- a/.idea/$CACHE_FILE$
+++ b/.idea/$CACHE_FILE$
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="NodePackageJsonFileManager">
+    <packageJsonPaths />
+  </component>
+</project>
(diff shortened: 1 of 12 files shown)


dg:
PHP reflection methods never returns `static` or `$this`

mabar:
I got Foo\Bar\mixed for class Foo\Bar\Baz class ([here](https://github.com/nette/schema/blob/v1.0.0/src/Schema/Helpers.php#L65)) so I expected that any type from phpdoc is also possible. I will write some tests for it

dg:
Oh, my comment was related to method `normalizeType()`, I did not notice `expandClassName()`.

mabar:
So should I make it expandClassName() only?
I accidentally commited .idea files, sorry for mess

dg:
I think the problem you are solving is in Schema, not in class Reflection. It is strictly related to the PHP (not to the phpDoc), so `mixed` for example is not built-in type and `$this` is not class name.

PR #205, opened 2019-12-02
Strings: throw exception on malformed UTF-8 in webalize() and toAscii()

VasekPurchart:
- bug fix
- BC break? no

Since `Strings::webalize()` can be called on user input it should be able to deal with anything or error in a predicable way. But when an malformed UTF-8 input is given currently it results in:

```
Nette\Utils\RegexpException was expected but got TypeError (strtr() expects parameter 1 to be string, null given)
```

(i guess this behaved differently before scrict types were added in [3.0.0](https://github.com/nette/utils/releases/tag/v3.0.0)).

Throwing an exception (the same exception as in other cases such as `Strings::replace()`) seems reasonable to me.

I am also not sure why the internal handling of PCRE is not used everywhere in calling `preg_*` is not used everywhere in the file, because there might be of course other problems too?

### DIFF
--- a/src/Utils/Strings.php
+++ b/src/Utils/Strings.php
@@ -142,7 +142,7 @@ class Strings
 			$transliterator = \Transliterator::create('Any-Latin; Latin-ASCII');
 		}
 
-		$s = preg_replace('#[^\x09\x0A\x0D\x20-\x7E\xA0-\x{2FF}\x{370}-\x{10FFFF}]#u', '', $s);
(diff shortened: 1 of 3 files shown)


VasekPurchart:
Travis failures seem to be unrelated to this PR.

dg:
Thanks

VasekPurchart:
@dg Thanks for merging. As for the other places in Strings where the native `preg_` functions are called - do you want them changed as well?

dg:
I did it 6339ca256bb7aa1a5f44d8b2281ed077f0eb8287

VasekPurchart:
Great, thanks 👍

PR #210, opened 2019-12-16
Strings: Fix boolean return of Normalizer.

janbarasek:
- bug fix
- BC break? yes

In case of `Strings::normalize("... \xbee m\xc3\xa1\xc5\xa1 ...");` normalizer throw error, because `\Normalizer::normalize` can return `false` in case of invalid string.

Now normalized string will be used only in case of valid return. In other cases keep original.

Snímek obrazovky 2019-12-16 v 14 31 24

Thanks.


dg:
Thanks

PR #213, opened 2020-02-05
Normalizer: Add support for fix malformedData

janbarasek:
- new feature
- BC break? yes

Add back compatibility support and parameter for fix broken string for example from database:

Current behavior:

Snímek obrazovky 2020-02-05 v 18 48 44

Thanks.


JanTvrdik:
* `normalize()` requires valid UTF-8 string as input (as documented)
* there is already `fixEncoding()` method

PR #214, opened 2020-02-17
Image: Return sizes must be integer only.

janbarasek:
- bug fix
- BC break? yes

I think return sizes from `calculateCutout` must be `int` only.

In case of `float` Image throw `TypeError`:

Snímek obrazovky 2020-02-17 v 14 51 34

Thansk.

### DIFF
--- a/src/Utils/Image.php
+++ b/src/Utils/Image.php
@@ -428,7 +428,7 @@ class Image
 		}
 		$newWidth = min($newWidth, $srcWidth - $left);
 		$newHeight = min($newHeight, $srcHeight - $top);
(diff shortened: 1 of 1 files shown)


dg:
I think you're hiding a usercode mistake. crop() accepts only int|string.

janbarasek:
@dg Ok, I understand.

I think `float` should be converted to `int`. What do you think about?

Because `float` is very common type, for example by `round($width / 2)` in user code.

dg:
But it creates inconsistency. PHP does not convert float to int in strict mode.

PR #217, opened 2020-03-15
Fixed type issue in crop function

zingon:
* bug fix
* BC break? no

When type of width and height is string without "%" char, line 392 and 393 throws "must be of the type integer, string given". (Happened with webchemistry/images img macro with exact resize).

### DIFF
--- a/src/Utils/Strings.php
+++ b/src/Utils/Strings.php
@@ -141,9 +141,7 @@ class Strings
 	 */
 	public static function toAscii(string $s): string
 	{
-		if (!extension_loaded('iconv')) {
-			throw new Nette\NotSupportedException(__METHOD__ . '() requires ICONV extension that is not loaded.');
-		}
+		$iconv = defined('ICONV_IMPL') ? ICONV_IMPL : null;
 		static $transliterator = null;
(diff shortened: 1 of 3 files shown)


src/Utils/Strings.php
```diff
@@ -147,24 +152,26 @@ public static function toAscii(string $s): string
 
 		// transliteration (by Transliterator and iconv) is not optimal, replace some characters directly
 		$s = strtr($s, ["\u{201E}" => '"', "\u{201C}" => '"', "\u{201D}" => '"', "\u{201A}" => "'", "\u{2018}" => "'", "\u{2019}" => "'", "\u{B0}" => '^', "\u{42F}" => 'Ya', "\u{44F}" => 'ya', "\u{42E}" => 'Yu', "\u{44E}" => 'yu']); // „ “ ” ‚ ‘ ’ ° Я я Ю ю
-		if (ICONV_IMPL === 'glibc') {
+		if ($iconv !== 'libiconv') {
```
adaamz, line 155: What about opposite comparison?
```suggestion
		if ($iconv === null) {
```

KminekMatej:
+1 for quick release - its currently blocking my autobuild flow

PR #221, opened 2020-05-15
Added Nette\Utils\Floats helper

mesour:
- new feature
- BC break? no
- doc PR: https://github.com/nette/docs/pull/848

---

Added support for comparing floats

The original class was created by @fabik.

I've added tests, and some fixes in the method `compare`.

### DIFF
--- a/src/Utils/Floats.php
+++ b/src/Utils/Floats.php
@@ -1,0 +1,109 @@
+<?php
+
+/**
+ * This file is part of the Nette Framework (https://nette.org)
+ * Copyright (c) 2004 David Grudl (https://davidgrudl.com)
+ */
+
+declare(strict_types=1);
+
+namespace Nette\Utils;
+
+use Nette;
+
+
+/**
+ * Floats compare tools library.
+ */
+class Floats
+{
+	use Nette\StaticClass;
+
+	/** @var float */
+	private const EPSILON = 1e-10;
+
+
+	public static function isZero(float $value): bool
+	{
+		return abs($value) < self::EPSILON;
+	}
+
+
+	public static function isInteger(float $value): bool
+	{
+		return abs(round($value) - $value) < self::EPSILON;
+	}
+
+
+	/**
+	 * Compare two floats. If $a < $b it returns -1, if they are equal it returns 0 and if $a > $b it returns 1
+	 * @throws \LogicException if one of parameters is NAN
+	 */
+	public static function compare(float $a, float $b): int
+	{
+		if (is_nan($a) || is_nan($b)) {
+			throw new \LogicException('Trying to compare NAN');
+
+		} elseif (!is_finite($a) && !is_finite($b) && $a === $b) {
+			return 0;
+
+		} else {
+			$diff = abs($a - $b);
+			if (($diff < self::EPSILON || ($diff / max(abs($a), abs($b)) < self::EPSILON))) {
+				return 0;
+			}
+		}
+
+		return $a < $b ? -1 : 1;
+	}
+
+
+	/**
+	 * Returns true if $a = $b
+	 * @throws \LogicException if one or both parameters are NAN
+	 */
+	public static function areEqual(float $a, float $b): bool
+	{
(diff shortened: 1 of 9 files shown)


src/Utils/Numbers.php
```diff
@@ -0,0 +1,98 @@
...
+
+namespace Nette\Utils;
+
+use Nette;
+
+
+/**
+ * Floats tools library.
+ */
+class Numbers
```
JanTvrdik, line 18: Should be called `Floats`.

src/Utils/Numbers.php
```diff
@@ -0,0 +1,98 @@
...
+
+	public static function isZero(float $value): bool
+	{
+		return abs($value) < self::$epsilon;
+	}
+
+
+	public static function isInteger(float $value): bool
+	{
+		return abs(floor($value) - $value) < self::$epsilon;
```
JanTvrdik, line 34: imho `round` must be used instead of `floor`

src/Utils/Numbers.php
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * Floats tools library.
+ */
+class Numbers
+{
+	use Nette\StaticClass;
+
+	/** @var float */
+	private static $epsilon = 1e-9;
```
JanTvrdik, line 23: should be constant

src/Utils/Numbers.php
```diff
@@ -0,0 +1,98 @@
...
+
+	public static function isInteger(float $value): bool
+	{
+		return abs(floor($value) - $value) < self::$epsilon;
+	}
+
+
+	public static function divide(float $dividend, float $divisor): float
+	{
+		if (self::isZero($divisor)) {
```
JanTvrdik, line 40: Not sure how useful this is. Division by zero is well defined for floats. If you don't want to get `+Inf`, `-Inf` or `Nan` as a result, you would need to check more things.

src/Utils/Numbers.php
```diff
@@ -0,0 +1,98 @@
...
+			}
+			throw new DivisionByZeroException('Trying to get a percentage of the total of 0.');
+		}
+		return ($part / $total) * 100;
+	}
+
+
+	public static function compare(float $a, float $b): int
+	{
+		if (abs($a - $b) < self::$epsilon) {
```
JanTvrdik, line 62: It's not that simple https://github.com/nette/tester/pull/68

src/Utils/Numbers.php
mesour: agree

src/Utils/Numbers.php
mesour: agree

src/Utils/Numbers.php
mesour: I don't know how much, but I did not encounter this problem. And I think, here is better to check at least zero than nothing.

If you have an example that would return NAN or INF, feel free to send it and try to fix it in that method.

src/Utils/Numbers.php
mesour: Upgraded. Now it works too with INF and NAN.

src/Utils/Numbers.php
JanTvrdik: * `Nan / 3.0`
* `Inf` / 4.0

src/Utils/Floats.php
```diff
@@ -0,0 +1,85 @@
...
+	private const EPSILON = 1e-10;
+
+
+	public static function isZero(float $value): bool
+	{
+		return abs($value) < self::EPSILON;
+	}
+
+
+	public static function isInteger(float $value): bool
```
janbarasek, line 32: Epsilon should be user-defined. :)

mesour: Now it is public static property :)

mesour: Changed back to constant. If it turns out why it should be user-defined, then we make an adjustment.

dg:
Great, thanks

PR #224, opened 2020-06-01
Strings: Add range() function with UTF-8 support.

janbarasek:
- new feature?
- BC break? no

Added new function `range()` for safe generating of characters interval.

It can be used in [Random](https://github.com/nette/utils/blob/master/src/Utils/Random.php#L28) helper for support intervals like `'0-9a-zA-Zá-ž'` (Czech `á` - `ž`).

Thanks.

### DIFF
--- a/tests/Utils/Strings.range().phpt
+++ b/tests/Utils/Strings.range().phpt
@@ -1,0 +1,43 @@
+<?php
+
+/**
+ * Test: Nette\Utils\Strings::range()
+ */
+
+declare(strict_types=1);
+
+
+use Nette\Utils\Strings;
+use Tester\Assert;
+
+
+require __DIR__ . '/../bootstrap.php';
+
+
+// Same input
+Assert::same(['a'], Strings::range('a', 'a'));
+
+// EN characters
+Assert::same(['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z'], Strings::range('a', 'z'));
+Assert::same(['A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z'], Strings::range('A', 'Z'));
+
+// Numbers
+Assert::same(['1', '2', '3', '4', '5', '6', '7', '8', '9'], Strings::range('1', '9'));
+Assert::same(['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'], Strings::range('0', '9'));
+Assert::same(['0', '1'], Strings::range('0', '1'));
+
+// Empty input
+Assert::exception(function () {
+	Strings::range('', '');
+}, \Nette\InvalidStateException::class);
+
+Assert::exception(function () {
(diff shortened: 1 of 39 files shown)


janbarasek:
@dg Some tests added for clear cases.

I am not sure what is best behavior for input like `range('á', 'ž')`. We can support all alphabets or most common characters only (stored in array or string)?

tests/Utils/Strings.range().phpt
```diff
@@ -0,0 +1,46 @@
...
+Assert::exception(function () {
+	Strings::range('', 'a');
+}, \Nette\InvalidStateException::class);
+
+Assert::exception(function () {
+	Strings::range('a', '');
+}, \Nette\InvalidStateException::class);
+
+// Special characters
+Assert::same(['á', 'â', 'ã', 'ä', 'å', 'æ', 'ç', 'è', 'é', 'ê', 'ë', 'ì', 'í', 'î', 'ï', 'ð', 'ñ', 'ò', 'ó', 'ô', 'õ', 'ö', '÷', 'ø', 'ù', 'ú', 'û', 'ü', 'ý', 'þ', 'ÿ', 'Ā', 'ā', 'Ă', 'ă', 'Ą', 'ą', 'Ć', 'ć', 'Ĉ', 'ĉ', 'Ċ', 'ċ', 'Č', 'č', 'Ď', 'ď', 'Đ', 'đ', 'Ē', 'ē', 'Ĕ', 'ĕ', 'Ė', 'ė', 'Ę', 'ę', 'Ě', 'ě', 'Ĝ', 'ĝ', 'Ğ', 'ğ', 'Ġ', 'ġ', 'Ģ', 'ģ', 'Ĥ', 'ĥ', 'Ħ', 'ħ', 'Ĩ', 'ĩ', 'Ī', 'ī', 'Ĭ', 'ĭ', 'Į', 'į', 'İ', 'ı', 'Ĳ', 'ĳ', 'Ĵ', 'ĵ', 'Ķ', 'ķ', 'ĸ', 'Ĺ', 'ĺ', 'Ļ', 'ļ', 'Ľ', 'ľ', 'Ŀ', 'ŀ', 'Ł', 'ł', 'Ń', 'ń', 'Ņ', 'ņ', 'Ň', 'ň', 'ŉ', 'Ŋ', 'ŋ', 'Ō', 'ō', 'Ŏ', 'ŏ', 'Ő', 'ő', 'Œ', 'œ', 'Ŕ', 'ŕ', 'Ŗ', 'ŗ', 'Ř', 'ř', 'Ś', 'ś', 'Ŝ', 'ŝ', 'Ş', 'ş', 'Š', 'š', 'Ţ', 'ţ', 'Ť', 'ť', 'Ŧ', 'ŧ', 'Ũ', 'ũ', 'Ū', 'ū', 'Ŭ', 'ŭ', 'Ů', 'ů', 'Ű', 'ű', 'Ų', 'ų', 'Ŵ', 'ŵ', 'Ŷ', 'ŷ', 'Ÿ', 'Ź', 'ź', 'Ż', 'ż', 'Ž', 'ž'], Strings::range('á', 'ž'));
```
dg, line 46: I think uppercase letters in range `'á..ž'` look unexpected.

dg: If it has to return all accented characters, there is missing `à`, because one would have to know that it should be called as `range('à', 'ž')`. (And maybe there are other characters missing like `ȁ`). It seems quite controversial to me.

JanTvrdik:
I don't think this is a good idea. Unicode ranges are rarely usable. It's not sorted logically like alphabet where the order is well defined.

PR #225, opened 2020-06-07
Validators::isEmail() should support localhost domain for debug

janbarasek:
- new feature
- BC break? yes

I think domain `localhost` should be supported in e-mail address for local debugging.

By the way, HTML5 input validation `<input type="email" required />` allows `localhost` as domain.


dg:
This is simply not a valid email address.

janbarasek:
Ok, sorry.

PR #226, opened 2020-06-09
FileSystem::createDir(): Added $mode to exception.

janbarasek:
- new feature
- BC break? no

For better debugging.

### DIFF
--- a/src/Utils/FileSystem.php
+++ b/src/Utils/FileSystem.php
@@ -135,7 +135,7 @@ final class FileSystem
 	{
 		static::createDir(dirname($file));
 		if (@file_put_contents($file, $content) === false) { // @ is escalated to exception
-			throw new Nette\IOException("Unable to write file '$file' with mode. " . Helpers::getLastError());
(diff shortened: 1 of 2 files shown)


src/Utils/FileSystem.php
```diff
@@ -135,7 +135,7 @@ public static function write(string $file, string $content, ?int $mode = 0666):
 	{
 		static::createDir(dirname($file));
 		if (@file_put_contents($file, $content) === false) { // @ is escalated to exception
-			throw new Nette\IOException("Unable to write file '$file'. " . Helpers::getLastError());
+			throw new Nette\IOException("Unable to write file '$file' with mode " . decoct($mode) . '. ' . Helpers::getLastError());
```
dg, line 138: $mode is not realated to file_put_contents()

janbarasek: I will fix it, sorry.

janbarasek: Fixed.

PR #227, opened 2020-06-23
Fix ICONV_IMPL constant

janfejtek:
- bug fix
- BC break? no

Constant ICONV_IMPL can contain quotes on some systems.
I could not find any official description why that is, same problem was reported here for example https://github.com/techdivision/import/pull/173


dg:
Thanks

PR #232, opened 2020-08-11
Webalize: Autoremove HTML &nbsp; and HTML tags from user input.

janbarasek:
- new feature
- BC break? yes

I think in some special cases `Strings::webalize()` should rewrite user input and ignore some tags.

For example my client want rewrite string (this is real example): `Ibalgin&nbsp;<sup>®</sup>&nbsp;Neco`. This feature return more clean output: `ibalgin-r-neco`.

Thanks.

### DIFF
--- a/.github/funding.yml
+++ b/.github/funding.yml
@@ -1,0 +1,2 @@
+github: dg
+custom: "https://nette.org/donate"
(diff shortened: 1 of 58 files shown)


dg:
There is Html::htmlToText for this.

janbarasek:
Ok...

PR #237, opened 2020-11-23
SmartObject: __set() and __unset() compatibility with Latte Strict trait

janbarasek:
- bug fix
- BC break? yes

Latte trait `Strict` should be compatible with `SmartObject`, because some packages use `SmartObject` inside `MacroSet`.

Snímek obrazovky 2020-11-23 v 20 03 51


dg:
IMHO these packages should be fixed. (Maybe duplicate to nette/latte#235)

PR #238, opened 2020-11-23
Html: Codestyle fix (broken test)

janbarasek:
- bug fix
- BC break? no

Just codestyle fix:

Snímek obrazovky 2020-11-23 v 20 08 44


dg:
This is fixed bug in slevomat/coding-style

Issue #239, opened 2020-11-26
German umlauts are converted incorrectly with "toAscii" (webalize) method


Zrnik:
Version: i think every version

### Bug Description
Method `toAscii` (and thus `webalize`) is incorrectly converting german umlauts.

### Reproduce

```php
<?php declare(strict_types=1);

include __DIR__ . '/src/Utils/StaticClass.php';
include __DIR__ . '/src/Utils/Callback.php';
include __DIR__ . '/src/Utils/Strings.php';

$ok = "✔️";
$err = "❌";

$CorrectConversions = [
    "Jägermeister" => "jaegermeister",
    "Würth" => "wuerth",
    "Löwenbräu" => "loewenbraeu",
    "Straße" => "strasse"
];

foreach ($CorrectConversions as $toConvert => $correctConversion) {
    $converted = \Nette\Utils\Strings::webalize($toConvert);
    $isCorrect = $converted === $correctConversion;
    echo $toConvert . " (".$correctConversion.") => " . ($isCorrect ? $ok : $err) . " " . $converted . PHP_EOL;
}
```
Result:
```text
Jägermeister (jaegermeister) => ❌ jagermeister
Würth (wuerth) => ❌ wurth
Löwenbräu (loewenbraeu) => ❌ lowenbrau
Straße (strasse) => ✔️ strasse
```

### Expected Behavior
```text
Jägermeister (jaegermeister) => ✔️ jaegermeister
Würth (wuerth) => ✔️ wuerth
Löwenbräu (loewenbraeu) => ✔️ loewenbraeu
Straße (strasse) => ✔️ strasse
```

Zrnik:
#240

Zrnik:
❤️

PR #240, opened 2020-11-26
Strings::toAscii() fixed to correctly convert german umlauts (BC break)

Zrnik:
This also applies on Strings::webalize() method. 

- bug fix 
- BC break? yes
- Issue: #239


dg:
Thanks

PR #244, opened 2021-01-05
FileSystem: Add method makeWritable()

janbarasek:
- new feature
- BC break? no

Very often I need to write to a directory where I cannot guarantee the correctly set permissions in advance.

This feature can try to fix the permissions according to the settings, or throw an exception.

Directory permissions are set recursively.


JanTvrdik:
1) fix permission is bad name; it should be called `changePermissions` (with removal of default values for `$dirMode` and `$fileMode`) or `makeWritable` (and keep the default values)
2) i'm not sure if you need the `$recursive` parameter as everything in `FileSytem` is always recursive

janbarasek:
@JanTvrdik Fixed, thanks.

PR #246, opened 2021-01-07
Arrays::updateDiff()

milo:
- new feature
- BC break? no
- doc PR: will

Proposed `Arrays::updateDiff()` compares two associative arrays and returns such items from later one, which does not exist or differs from first one. By other words - which items have to be updated in first array, to be the same as second array.

As far as I know, there is no PHP function for that ([playground](https://3v4l.org/YW0CH)). For examle:
```php
$from = ['a' => null];
$to = ['a' => false];

# I didn't find a PHP function which returns
$diff = ['a' => false];
```

There are posible things to debate:
- function name
- arguments order
- recursive version
- objects comparing
- 3rd argument `callable $comparator = null`

### DIFF
--- a/src/Utils/Arrays.php
+++ b/src/Utils/Arrays.php
@@ -343,4 +343,19 @@ class Arrays
 	{
 		return key([$value => null]);
 	}
+
+
+	/**
+	 * Returns items required to synchronize associative array $from to array $to.
+	 */
+	public static function updateDiff(array $from, array $to): array
+	{
+		$diff = [];
+		foreach ($to as $k => $v) {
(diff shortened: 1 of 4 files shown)


JanTvrdik:
So it's like `array_diff_assoc` but with strict comparision?

milo:
Yes. Behaviour probably same as:
```php
return array_udiff_assoc($to, $from, function ($a, $b) {
	return $a === $b ? 0 : ($a > $b ? 1 : -1);  # spaceship cannot be used because null <=> false === 0
});
```
I write probably, because I'm not sure why yes/no comparator can return -1/0/1 and how result depends on it.
And the foreach loop is ~3 times faster.

PR #256, opened 2021-06-28
DateTime: Add shortcut now().

janbarasek:
- new feature
- BC break? no

Just feature for most common shortcut.


dg:
How is this different from `new DateTime`?

janbarasek:
The method is not magical because it expresses exactly what will be returned (evidence base practice). You can also override the method in your own implementation and, for example, handle time zones differently.

dg:
It is an unnecessary duplication of known functionality.

PR #260, opened 2021-07-29
FileSystem: fix createDir race condition

mabar:
- bug fix
- BC break? no

I recently started testing code in parallel and builds started failing with error [`Nette\IOException: Unable to create directory '/directory' with mode 777. File exists`](https://github.com/orisai/nette-console/runs/3191840764?check_suite_focus=true#step:15:20). File cannot be created because it exists. Apparently PHP does not clear file stat cache when `mkdir()` is called and we have to handle it ourself.

Note: `clearstatcache(true, $dir);` nor `clearstatcache()` are enough, for some reason only complete cache removal including all files and realpath cache works. Should not affect performance much because it is run only in case of this race condition. `clearstatcache(true)` is the only variant that with 10 runs didn't fail for me, other variants failed with 2 runs max.

Funny thing: phpunit, symfony, laravel and others do not handle that properly neither.

### DIFF
--- a/src/Utils/FileSystem.php
+++ b/src/Utils/FileSystem.php
@@ -25,8 +25,11 @@ final class FileSystem
 	 */
 	public static function createDir(string $dir, int $mode = 0777): void
 	{
-		if (!is_dir($dir) && !@mkdir($dir, $mode, true) && !is_dir($dir)) { // @ - dir may already exist
-			throw new Nette\IOException("Unable to create directory '$dir' with mode " . decoct($mode) . '. ' . Helpers::getLastError());
+		if (!is_dir($dir) && !@mkdir($dir, $mode, true)) { // @ - dir may already exist
+			clearstatcache(true, $dir);
+			if (!is_dir($dir)) {
+				throw new Nette\IOException("Unable to create directory '$dir' with mode " . decoct($mode) . '. ' . Helpers::getLastError());
+			}
 		}
 	}
 
--- a/src/Utils/FileSystem.php
+++ b/src/Utils/FileSystem.php
@@ -25,11 +25,9 @@ final class FileSystem
 	 */
 	public static function createDir(string $dir, int $mode = 0777): void
 	{
-		if (!is_dir($dir) && !@mkdir($dir, $mode, true)) { // @ - dir may already exist
-			clearstatcache(true, $dir);
-			if (!is_dir($dir)) {
-				throw new Nette\IOException("Unable to create directory '$dir' with mode " . decoct($mode) . '. ' . Helpers::getLastError());
-			}
+		// @ - dir may already exist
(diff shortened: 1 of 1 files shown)


milo:
Do you create the single level depth dir or deeper one?

mabar:
3 levels

milo:
I don't think that stat cache is the reason. `mkdir()` is not in the affected function list https://www.php.net/manual/en/function.clearstatcache.php

If `mkdir()` creates `a/b/c/d/e/f` it may looks like:
1st thread - does `a/b/c` exists? No.
2nd thread - does `a/b/c` exists? No.
1st thread - create `a/b/c`. OK.
2nd thread - create `a/b/c`. Failed: already exists

mabar:
mkdir() is not affected, but is_dir() is.

I think what is happening is that only after successful call to mkdir() is cache used by is_dir() updated. And because it failed we have to drop cache ourself.

dg:
> You should also note that PHP doesn't cache information about non-existent files. https://www.php.net/manual/en/function.clearstatcache.php

First condition `!is_dir($dir)` is therefore not cached.

mabar:
I tried to dig deeper into that problem and even from reading php-src I can't really tell what's wrong.

For some reason, both `mkdir()` and second `is_dir()` return `false`. I suspect that `mkdir()` returns `false` not because dir already exists, but because it's *just being created* in the other thread and second `is_dir()` returns false because directory is *still being created*.

So I replaced `clearstatcache()` with `usleep(1000)` and even after 64 runs the failure does not happen anymore. File stat was not really a problem, clearing its cache just provided necessary delay. Before the fix at least each 10th run failed.

I have no idea whether the issue is in php, operating system or particular filesystem since I am not able to reproduce it with none of my devices nor with a server, only with github actions.
Any help with debugging and reporting issue is appreciated. Or maybe we can just try combine second `is_dir` with `usleep()` and repeat it few times until it succeeds or run out of attempts.

milo:
I hit this issue years ago in Tester. And IIRC, the issue exists only in non thread safe (or thread safe) PHP version and it is related to a filesystem, because the filesystem itself is not capable to atomic deep directory creation.

Due to that, it cannot be safely implemented even in the loop, because another thread can remove upper directory before final directory leaf is created.

JanTvrdik:
also note that mkdir behavior is changing in PHP 8.1 (or maybe 8.2?) https://github.com/php/php-src/pull/7383

mabar:
> also note that mkdir behavior is changing in PHP 8.1 (or maybe 8.2?) [php/php-src#7383](https://github.com/php/php-src/pull/7383)

Great! I tried to run few hundred jobs with PHP 8.1.0-dev (current master) and none failed. 
https://github.com/orisai/nette-console/runs/3394146735?check_suite_focus=true

PR #263, opened 2021-08-09
Paginator: add firstItemOnPage & lastItemOnPage [Closes #262]

### LINKED ISSUE #262: paginator: add methods for first and last record on page
- Explain your intentions
Add two simple methods to the paginator, that could be useful for the user of the paginator: `firstRecordOnPage`, `lastRecordOnPage`. It would allow to display current record range on page in pagination visualization when needed (eg. "records 20 - 27 of 27").

- It's up to you to make a strong case to convince the project's developers of the merits of this feature
These methods provide quite simple calculations, that are based on data already known in paginator. When one wants to display this in application, it must be calculated all over again. It seems logic to have all the pagination math here.

I would prepare PR in case you found this request beneficial. Otherwise feel free to close it.

khorsky:
- new feature? #262  
- BC break? no
- doc PR: nette/docs#??? I'm gonna prepare it in a minute...

### DIFF
--- a/src/Utils/Paginator.php
+++ b/src/Utils/Paginator.php
@@ -18,6 +18,8 @@ use Nette;
  * @property   int $page
  * @property-read int $firstPage
  * @property-read int|null $lastPage
+ * @property-read int $firstItemOnPage
+ * @property-read int $lastItemOnPage
  * @property   int $base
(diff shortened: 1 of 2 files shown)


dg:
Thanks!

PR #264, opened 2021-09-13
Strings: Add removeEmoji().

janbarasek:
- new feature
- BC break? no

Very useful feature for removing emoji, for example before saving to DB:

Snímek obrazovky 2021-09-13 v 11 57 52

Inspired by https://stackoverflow.com/questions/61481567/remove-emojis-from-string.


mabar:
Such feature should ideally be in sync with unicode.org list of emojis via automation. Otherwise it's highly unreliable because it does not handle all ranges and would be hard to keep up to date with ever growing list of emojis.

Also note it will help only with removing emojis itself, not fixing collation issues in your database because unsupported unicode ranges include way more than just emojis. Real fix of database collation issues is in upgrading to newer collations, e.g. `utf8mb4_0900_ai_ci` which supports Unicode 9.0 (unfortunatelly it does not seem like unicode 10.0 to 13.0 are supported by mysql in any way) or `utf8mb4_unicode_520_ci ` when other one is not supported.

https://unicode.org/Public/emoji/13.1/emoji-sequences.txt

PR #269, opened 2021-10-12
Validators::isEmail: Add check for DNS nameservers.

janbarasek:
- new feature
- BC break? no

E-mail should be valid only in case of domain name use valid `MX` DNS record.

Examples:

```php
Validators::isEmail('jan@barasek.com'); // true
Validators::isEmail('jan@aaadfergrtgeaaa.com'); // true
Validators::isEmail('jan@aaadfergrtgeaaa.com', true); // false
```

### DIFF
--- a/src/Utils/Validators.php
+++ b/src/Utils/Validators.php
@@ -294,11 +294,11 @@ class Validators
 	/**
 	 * Checks if the value is a valid email address. It does not verify that the domain actually exists, only the syntax is verified.
 	 */
-	public static function isEmail(string $value): bool
+	public static function isEmail(string $value, bool $checkDns = false): bool
 	{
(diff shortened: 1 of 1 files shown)


dg:
@janbarasek can you update PR please?

nechutny:
There was forum post about this 8 years ago https://forum.nette.org/en/1293-mail-adress-validation-mx-records and it contains good points which needs to be solved.

mabar:
> You don't need MX record. A, or AAAA is sufficient to deliver e-mails.

I just encountered case when it was not. Customer had domain with only MX for sending emails and no A or AAAA records.

milo:
And the conclusion?

PR #276, opened 2022-01-10
Validators: isNumeric support string number '1.'

h4kuna:
- bug fix
- BC break? no

The goal of this change is same behavior for float and string number.

### Old behavior
```php
Validators::isNumeric(1.); // float number true
Validators::isNumeric('1.'); // string number false
```
### New behavior
```php
Validators::isNumeric(1.); // float number true
Validators::isNumeric('1.'); // string number true
```

And I wrote test for to keep back compatibility. How you can see in three commits.


dg:
Thanks

PR #279, opened 2022-01-31
Delete: During the deletion of a folder, additional files may be added.

janbarasek:
- bug fix
- BC break? yes

New files may be created by another process while the directory is being deleted. The deletion will then fail. The new implementation repeatedly checks the files in the directory.

Demo:

Snímek obrazovky 2022-01-31 v 13 22 31

Snímek obrazovky 2022-01-31 v 13 32 52

Thanks.

### DIFF
--- a/src/Utils/FileSystem.php
+++ b/src/Utils/FileSystem.php
@@ -96,9 +96,13 @@ final class FileSystem
 				));
 			}
 		} elseif (is_dir($path)) {
-			foreach (new \FilesystemIterator($path) as $item) {
-				static::delete($item->getPathname());
-			}
+			$items = [];
+			do {
+				foreach ($items as $item) {
+					static::delete($item->getPathname());
+				}
+				$items = iterator_to_array(new \FilesystemIterator($path));
+			} while ($items !== []);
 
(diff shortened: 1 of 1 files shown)


JanTvrdik:
The error seems expected. You simply cannot delete a directory that is actively being written into.

janbarasek:
This means that, for example, the cache can never be removed if there are many parallel requests to the application.

How to safely delete the cache contents if there is an error?

For example, the `rm -fr temp/cache/*` command will always delete the contents safely and not cause an error to be thrown, like Nette.

JanTvrdik:
> This means that, for example, the cache can never be removed if there are many parallel requests to the application.

You can remove indivial files, just not the directory itself.

milo:
> For example, the `rm -fr temp/cache/*` command will always delete the contents safely and not cause an error to be thrown, like Nette.

That's because of `-f` option. It only mutes errors. Try `-r` only on rush directory. You get `Directory not empty`.

janbarasek:
So could Nette implement a `force` switch? There are certainly valid cases where this is useful.

PR #280, opened 2022-02-01
Add getIntlDateFormat method in Datetime class

petrparolek:
- new feature? yes
- BC break? no
- doc PR: nette/docs# I can prepare if PR will be merged

In Nette Utils I didn't find method which I could replace deprecated PHP function `strftime('%B', $datetime->getTimestamp()),` which I used to Czech translation of munth.

I found solution at https://stackoverflow.com/a/46275845/6346869

### DIFF
--- a/src/Utils/DateTime.php
+++ b/src/Utils/DateTime.php
@@ -144,4 +144,22 @@ class DateTime extends \DateTime implements \JsonSerializable
 		$dolly = clone $this;
 		return $modify ? $dolly->modify($modify) : $dolly;
 	}
+
+
+	public static function getIntlDateFormat(
+		$format = 'yyyy-MM-dd HH:mm:ss',
+		$time = null,
+		$locale = 'en_GB',
+		$timeZone = 'Europe/London'
+	): string {
+		if (empty($time)) {
+			$time = new \DateTime;
+		}
+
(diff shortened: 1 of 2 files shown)


petrparolek:
here are formats https://unicode-org.github.io/icu/userguide/format_parse/datetime/#date-field-symbol-table

stepapo:
For me this would be useful, but why is it static and not like `return $formatter->format($this);`?

JanTvrdik:
Instead of 

~~~php
echo Nette\Utils\DateTime::getIntlDateFormat(
	format: 'LLLL YYYY HH:mm:ss',
	time: new DateTimeImmutable('2021-02-01 15:47:21'),
	locale: 'cs_CZ',
	timeZone: 'Europe/Prague'
)
~~~

you can just use

~~~php
echo IntlDateFormatter::formatObject(
    datetime: new DateTimeImmutable('2021-02-01 15:47:21', new DateTimeZone('Europe/Prague')),
    format: 'LLLL YYYY HH:mm:ss',
    locale: 'cs_CZ',
);
~~~

petrparolek:
> For me this would be useful, but why is it static and not like `return $formatter->format($this);`?

thanks

dg:
I think it's better to use IntlDateFormatter::formatObject

petrparolek:
thanks @dg and @JanTvrdik for solution

PR #286, opened 2022-08-29
PHP 8.2 Support

driesvints:
- new feature?
- BC break? no

Depends on:
- [x] https://github.com/nette/tester/pull/440


driesvints:
@dg seems like some of these tests were already failing on master. The failing `gd` tests are because of the extension not being available on Ubuntu yet I think.

dg:
Done

driesvints:
Thanks!

PR #287, opened 2022-11-30
Strings: add base64UrlEncode() and base64UrlDecode()

milo:
- new feature
- BC break no
- doc PR: (will add into https://doc.nette.org/cs/utils/strings#toc-kodovani if accepted)

Base64 encoding is well known. It uses alphabet of 64 chars `A-Z a-z 0-9 + /` and padding char `=`. While the `A-Z a-z 0-9` chars are URL safe, the remaining `+ / =` are not. So there is a "Base 64 Encoding with URL and Filename Safe Alphabet" as mentioned in [RFC4648](https://datatracker.ietf.org/doc/html/rfc4648#page-7). This encoding replaces `+` by `-` and `/` by `_` and drops `=` padding.

The base64Url encoding is for example used by JSON Web Tokens (JWT), which are used for example in Open ID Connect protocol.

The implementation is based on [Appendix C of IETF draft](https://datatracker.ietf.org/doc/html/draft-ietf-jose-json-web-signature#appendix-C).


dg:
I didn't get it at all :) What does it mean that something is not url safe?

milo:
It is from mentioned RFC :) My opinion...

They wanted to transfer data via URL, but the `rawurlencode("+/=") === '%2B%2F%3D'` - these 3 chars from base64 alphabet have to be escaped and moreover, result is longer. So they brought `-_` replacements which are not escaped. From my point of view it is a bad idea to insert something directly into URL, the `rawurlencode()` should be always used.

I'm working with JSON Web Tokens on a 4th project this year and content of the token is `base64url` encoded. So I thought, that base64 decode/encode can be useful for other people. But there are no 👍 reactions so maybe I'm alone :)

PR #290, opened 2023-02-07
Improve type annotations

xificurk:
- new feature
- BC break? no
- doc PR: not needed

Add more specific type annotations to improve static analysis support.

### DIFF
--- a/src/Utils/ArrayList.php
+++ b/src/Utils/ArrayList.php
@@ -15,6 +15,8 @@ use Nette;
 /**
  * Provides the base class for a generic list (items can be accessed by index).
  * @template T
+ * @implements \IteratorAggregate<int, T>
+ * @implements \ArrayAccess<int, T>
  */
 class ArrayList implements \ArrayAccess, \Countable, \IteratorAggregate
 {
@@ -25,7 +27,7 @@ class ArrayList implements \ArrayAccess, \Countable, \IteratorAggregate
 
 	/**
 	 * Transforms array to ArrayList.
-	 * @param  array<T>  $array
+	 * @param  list<T>  $array
 	 */
 	public static function from(array $array): static
 	{
--- a/src/Utils/Arrays.php
+++ b/src/Utils/Arrays.php
@@ -225,6 +225,7 @@ class Arrays
 
 	/**
 	 * Checks if the array is indexed in ascending order of numeric keys from zero, a.k.a list.
+	 * @return ($value is list ? true : false)
 	 */
 	public static function isList(mixed $value): bool
 	{
--- a/src/Utils/Image.php
+++ b/src/Utils/Image.php
@@ -90,8 +90,8 @@ use Nette;
  * @method void stringUp($font, $x, $y, string $s, $col)
  * @method void trueColorToPalette(bool $dither, $ncolors)
  * @method array ttfText($size, $angle, $x, $y, $color, string $fontfile, string $text)
- * @property-read int $width
- * @property-read int $height
+ * @property-read positive-int $width
+ * @property-read positive-int $height
  * @property-read \GdImage $imageResource
  */
 class Image
@@ -220,6 +220,8 @@ class Image
 
 	/**
 	 * Creates a new true color image of the given dimensions. The default color is black.
+	 * @param positive-int $width
+	 * @param positive-int $height
 	 * @throws Nette\NotSupportedException if gd extension is not loaded
 	 */
 	public static function fromBlank(int $width, int $height, ?array $color = null): static
@@ -247,6 +249,7 @@ class Image
 
 	/**
 	 * Returns the type of image from file.
+	 * @return ImageType::*|null
 	 */
 	public static function detectTypeFromFile(string $file, &$width = null, &$height = null): ?int
 	{
@@ -257,6 +260,7 @@ class Image
 
 	/**
 	 * Returns the type of image from string.
+	 * @return ImageType::*|null
 	 */
 	public static function detectTypeFromString(string $s, &$width = null, &$height = null): ?int
 	{
@@ -267,6 +271,8 @@ class Image
 
 	/**
 	 * Returns the file extension for the given `ImageType::XXX` constant.
+	 * @param ImageType::* $type
+	 * @return value-of<self::Formats>
 	 */
 	public static function typeToExtension(int $type): string
 	{
@@ -280,6 +286,7 @@ class Image
 
 	/**
 	 * Returns the `ImageType::XXX` constant for given file extension.
+	 * @return ImageType::*
 	 */
 	public static function extensionToType(string $extension): int
 	{
@@ -295,6 +302,7 @@ class Image
 
 	/**
 	 * Returns the mime type for the given `ImageType::XXX` constant.
+	 * @param ImageType::* $type
 	 */
 	public static function typeToMimeType(int $type): string
 	{
@@ -314,6 +322,7 @@ class Image
 
 	/**
 	 * Returns image width.
+	 * @return positive-int
 	 */
 	public function getWidth(): int
 	{
@@ -323,6 +332,7 @@ class Image
 
 	/**
 	 * Returns image height.
+	 * @return positive-int
 	 */
 	public function getHeight(): int
 	{
@@ -351,7 +361,7 @@ class Image
 
 	/**
 	 * Scales an image. Width and height accept pixels or percent.
-	 * @param  self::OrSmaller|self::OrBigger|self::Stretch|self::Cover|self::ShrinkOnly  $mode
+	 * @param  int-mask-of<self::OrSmaller|self::OrBigger|self::Stretch|self::Cover|self::ShrinkOnly>  $mode
 	 */
 	public function resize(int|string|null $width, int|string|null $height, int $mode = self::OrSmaller): static
 	{
@@ -388,7 +398,7 @@ class Image
 
 	/**
 	 * Calculates dimensions of resized image. Width and height accept pixels or percent.
-	 * @param  self::OrSmaller|self::OrBigger|self::Stretch|self::Cover|self::ShrinkOnly  $mode
+	 * @param  int-mask-of<self::OrSmaller|self::OrBigger|self::Stretch|self::Cover|self::ShrinkOnly>  $mode
 	 */
 	public static function calculateSize(
 		int $srcWidth,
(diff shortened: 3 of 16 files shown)


src/Utils/Image.php
```diff
@@ -639,6 +653,7 @@ public function send(int $type = self::JPEG, ?int $quality = null): void
 	/**
 	 * Outputs image to browser or file.
 	 * @throws ImageException
+	 * @param self::JPEG|self::PNG|self::GIF|self::WEBP|self::AVIF|self::BMP $type
```
mabar, line 656: This type is repeated quite a lot, maybe type alias would be useful? https://phpstan.org/writing-php-code/phpdoc-types#local-type-aliases

xificurk: Yes, but I wanted to stay away from phpstan specific annotations, if possible.

Another alternative would be `key-of<self::Formats>`, but I'm not sure how well that is supported in the broader tooling.

mabar: Psalm and phpstan suppport each others annotations. Also phpstorm adds support for new phpstan and psalm annotations with each version.

Others probably burn when anything advanced is used...

src/Utils/ArrayList.php
```diff
@@ -15,6 +15,8 @@
 /**
  * Provides the base class for a generic list (items can be accessed by index).
  * @template T
+ * @implements \IteratorAggregate<int, T>
```
dg, line 18: Isn't that duplicate information?

src/Utils/Arrays.php
```diff
@@ -225,6 +225,7 @@ public static function flatten(array $array, bool $preserveKeys = false): array
 
 	/**
 	 * Checks if the array is indexed in ascending order of numeric keys from zero, a.k.a list.
+	 * @return ($value is list ? true : false)
```
dg, line 228: What's it good for?

src/Utils/Image.php
```diff
@@ -90,8 +90,8 @@
  * @method void stringUp($font, $x, $y, string $s, $col)
  * @method void trueColorToPalette(bool $dither, $ncolors)
  * @method array ttfText($size, $angle, $x, $y, $color, string $fontfile, string $text)
- * @property-read int $width
- * @property-read int $height
+ * @property-read positive-int $width
+ * @property-read positive-int $height
```
dg, line 94: What's it good for?

src/Utils/Image.php
```diff
@@ -267,6 +271,8 @@ public static function detectTypeFromString(string $s, &$width = null, &$height
 
 	/**
 	 * Returns the file extension for the given `Image::XXX` constant.
+	 * @param self::JPEG|self::PNG|self::GIF|self::WEBP|self::AVIF|self::BMP $type
+	 * @return value-of<self::Formats>
```
dg, line 275: What's `@return` annotation good for?

src/Utils/Image.php
```diff
@@ -280,6 +286,8 @@ public static function typeToExtension(int $type): string
 
 	/**
 	 * Returns the `Image::XXX` constant for given file extension.
+	 * @param value-of<self::Formats>|'jpg' $extension
+	 * @return self::JPEG|self::PNG|self::GIF|self::WEBP|self::AVIF|self::BMP
```
dg, line 290: What is `@param` good for?

src/Utils/Arrays.php
mabar: If variable is of type list it will always return true. Informs user that check is effectively useless.

But it should probably be `@return ($value is list ? true : bool)`

src/Utils/ArrayList.php
mabar: It is, but just for getIterator(). Implementing IteratorAggregate also passes generic type to Traversable. Without it Traversable type is mixed. Not sure if this specific case is problem, but generally speaking, generic interface should be always implemented.

Generic type from getIterator() should be removed, as it is already defined by IteratorAggregate

src/Utils/Image.php
mabar: Mostly for compatibility with other classes that already expect width and height to be >= 0. e.g. for database entity.

src/Utils/Image.php
mabar: Analysis reports if unsupported format is given. Instead of finding out later when `Unsupported file extension` is thrown

src/Utils/ArrayList.php
dg: Thanks for the explanation. Can I ask @xificurk  to remove the annotation from `getIterator()`?

src/Utils/Arrays.php
dg: I understand what it does, I ask what good it does :-) I just need an example.

src/Utils/Image.php
dg: Can you think of a real example where static analysis could detect what is being sent to this method and prevent exception?

dg: Can you think of a real example where static analysis could detect what is being sent to this method and prevent exception?

src/Utils/Image.php
dg: (I'm thinking that in nette/utils 4.1, which will require newer PHP, it can be replaced somehow with enum.)

src/Utils/Arrays.php
mabar: Here it is. `If condition is always true.`
https://phpstan.org/r/37d792a0-1fe7-488a-8818-15048810f6c0

src/Utils/Image.php
mabar: If you want class now, use enum-like class. Changing it safely to enum needs changing constants to cases, removing compatibility methods and deprecating static constructors
https://github.com/orisai/auth/blob/9e35c56fdab1871bdb903b693441574e23fe0b2a/src/Authentication/LogoutCode.php

src/Utils/Arrays.php
mabar: Sadly it is only half of the solution and it seems phpstan does not support assert for list. Even after isList() is given value array
https://phpstan.org/r/c26d6864-a46a-4d81-9fad-54dbc75d7e83

src/Utils/Image.php
mabar: Yeah, probably not. Unless someone has static list of supported extensions in app, it is gonna be just annoying.
But I would keep return type, narowing down return types does no harm

src/Utils/Arrays.php
enumag: Here is a better example. PHPStan does support asserts but the annotation needs to be prefixed.

https://phpstan.org/r/a7d26dd0-17a2-4cd7-abda-24e5d302d61d

mabar: @enumag Thank you! I didn't know unprefixed version is not recognized...

src/Utils/Image.php
dg: Yeah, return type is totally fine

src/Utils/Arrays.php
dg: @enumag `@return` is probably pointless then, isn't it?

src/Utils/Image.php
dg: It could perhaps be prearranged like this fe69e377b33862e73f2859beea95deb5c49cfa90

src/Utils/Image.php
xificurk: Good point, this would probably report more false positives than actual bugs. I'll remove it.

src/Utils/Arrays.php
enumag: @dg Yeah, it is indeed. I even checked Psalm. The return annotation doesn't make any difference for it either.

src/Utils/Image.php
xificurk: As any type (narrowing) hint, it's good to make code more readable and eliminate useless checks.

If you rely on static analysis and narrow down the types enough, [checks like this](https://github.com/nette/utils/pull/290/files#diff-94b775b0ed8e491ffeeccac27715ea411933e493574562c3ff790577661ab58fR233) become useless and can be safely eliminated.

Dimensions are often used in various calculations. Knowing that those are actually positive integers, implies e.g.:
- their ratio is well defined and also positive number
- the area (number of pixels in image) is positive integer

src/Utils/Image.php
xificurk: For me the most important thing out of this is specifying that the output extension is non-empty string, which is useful when you want to use the extension further. I guess `@return non-empty-string` could cover most of the use-cases, but why shouldn't we make the return type as narrow as possible?

dg: Now, as a return annotation, this should probably be `ImageType::*`, right?

dg:
I would like to release version 4.0.1 soon, we can finish this somehow @xificurk ?

xificurk:
@dg I'll take a look today/tomorrow.

src/Utils/Image.php
xificurk: No, because it returns one of the string values contained in `self::Formats`, not `ImageType::*` value.

src/Utils/Arrays.php
xificurk: @mabar @enumag Why do you think the return type should be `@return ($value is list ? true : **bool**)`? Can you give me example of value type which is not list, but the method will return true?

This change was actually the reason, why the examples you've posted seemingly did not work. The correct conditional return annotation is enough for phpstan to correctly detect always true/false conditions & perform type narrowing.

https://phpstan.org/r/c3f4cdc0-3417-42d8-9e5f-24f1c722b01c

enumag: @xificurk Because $value can be a list without PHPStan being aware of it.

xificurk: @enumag Can you create example in phpstan playground?

btw, isn't it exactly the case of the last if-else block in [my example](https://phpstan.org/r/c3f4cdc0-3417-42d8-9e5f-24f1c722b01c)? PHPStan can't know if the value is list or not, but if it is, the `isList()` method will always return true, otherwise false.

enumag: I tried a few things and yeah you're right, false is enough there. PHPStan does treat unknown array as a possible list.

dg:
Thanks!

PR #295, opened 2023-05-16
Finder: Fix symlink handling

nufue:
- bug fix / new feature?   bug fix
- BC break? possibly (changes behavior in nette/utils:4 to be compatible with behavior in nette/finder:2)
- doc PR: not necessary

### Bug Description
Finder (v2) did use `isFile()` or `isDir()` to determine type of `SplFileInfo` ([link](https://github.com/nette/finder/blob/991aefb42860abeab8e003970c3809a9d83cb932/src/Utils/Finder.php#L101)).

After migration to `nette/utils`, it uses `SplFileInfo::getType()` ([link](https://github.com/nette/utils/blob/0a94bd77ff1775e039aa4416abc9807d12c198ba/src/Utils/Finder.php#L385)).

There is a slight difference between the two approaches regarding handling of symlinks - if `SplFileInfo` is a symlink to a file, `isFile()` returns `true` (although it is not documented - see [comment](https://www.php.net/manual/en/splfileinfo.isfile.php#123963)), while `getType()` returns `link` (and thus fails `=== 'file'` comparison).

This leads to change of behavior in nette/utils v4 when Finder only returns real files, not symlinks and there is no way to return to the previous behaviour. 

### Steps To Reproduce
1. create a symlink to file in a directory (`mkdir ./test && touch ./file.txt && cd ./test && ln -s ../file.txt`)
2. `Finder::findFiles('*')->in('./test')` returns empty array


dg:
Thanks

PR #296, opened 2023-06-10
Utilize PhpToken::tokenize() when available

staabm:
- bug fix / new feature?   
- BC break? no
- doc PR: nette/docs#???  

I can see Rector to bottleneck on `token_get_all`.

grafik

do you think we could backport this change to `PhpToken::tokenize` to 3.x so libs which need php7 compat can still benefit from the memory/perf improvements of `PhpToken::tokenize` when run on php8?

----

php7 compatible backport of https://github.com/nette/utils/commit/8e2e71cd702a59ba4e786bb031135583e2f1ef23

### DIFF
--- a/src/Utils/Reflection.php
+++ b/src/Utils/Reflection.php
@@ -309,8 +309,14 @@ final class Reflection
 	 */
 	private static function parseUseStatements(string $code, ?string $forClass = null): array
 	{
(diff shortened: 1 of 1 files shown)


staabm:
did some measures locally and it seems the differences are negligible

PR #298, opened 2023-08-05
Arrays: add generic type declaration for some(), every() and map()

JanTvrdik:
Tests: https://phpstan.org/r/85fa16b9-d5f7-4315-b0f1-e95d8a328ac7


dg:
OT: Do you know how to fix this? 
image

enumag:
Use object instead of \stdClass.

https://phpstan.org/writing-php-code/phpdoc-types#object-shapes

PR #299, opened 2023-09-07
Avoid deprecation notice when using utf8_decode 

KminekMatej:
- bug fix
- BC break? no

PHP 8.2 throws deprecation notice on `utf8_encode` and `utf8_decode` functions. `Strings::length` used `utf8_decode` to determine correct length of string - using `mb_strlen`, or natively using `utf8_decode`. I refactored this function to use `mbstring` or `iconv` extensions to determine correct length.

User can no longer run `Strings::length()` function without at least one of these extensions, however since they are super common extensions I dont see this as s problem nowadays


dg:
I did it a little bit differently.

KminekMatej:
Even better

PR #301, opened 2023-09-13
DateTime: method setTime() allow null value for keep current value

h4kuna:
- new feature?  yes
- BC break? no
- doc PR: nette/docs#???  if you agree with theses changes I will write.

I keep default values like a parent. The back compatibility is ok.

You can more simply modify part of time. For example you want reset microseconds

```php
dump((new Nette\Utils\DateTime())->setTime(null, null, null, 0));
```

More examples are in test.

### DIFF
--- a/src/Utils/DateTime.php
+++ b/src/Utils/DateTime.php
@@ -111,6 +111,22 @@ class DateTime extends \DateTime implements \JsonSerializable
 	}
 
 
+	public function setTime(
+		?int $hour,
+		?int $minute,
+		?int $second = 0,
+		?int $microsecond = 0,
+	): static
+	{
+		return parent::setTime(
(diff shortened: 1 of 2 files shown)


dg:
This is pretty tricky. To make it sensible, `setTime(microsecond: 0)` would have to work, because the notation with many zeros and null is not good. But if you implement it that way, the whole thing can behave in surprising ways.

h4kuna:
Yes you have right, all parameters must have default value null. The example `setTime(microsecond: 0)` will work right. But I keep back compatibility. Because if all parameters will be null, this exists code will to work bad. `setTime(1, 1)`. 

The way where are all parameters null I like it more too.

h4kuna:
ok

PR #302, opened 2023-09-13
DateTimeFormat: support quick access properties

h4kuna:
- bug fix / new feature?  yes
- BC break? no
- doc PR: nette/docs#???   if you agree with theses changes I will write.

I add trait DateTimeFormat what extends class DateTime, with quick access to frequent date and time values with right type.

I used SmartObject and added read-only properties for DateTime.

### Example

```php
use Nette\Utils\DateTime;

$now = new DateTime();
dump($now->year === 2023); // true
dump($now->year === '2023'); // false, 2023 is string
dump($now->month === 5);
dump($now->day === 13);
```


h4kuna:
If you don't like it, let's close it.

PR #303, opened 2023-10-23
Function ReflectionType::__toString() is deprecated

tacoberu:
- bug fix
- BC break? no

In php7.4 version it Function ReflectionType::__toString() is deprecated. It breaks my projects that I would like to keep on php7.4.


dg:
According to travis.yml, Nette Utils 2.4 is compatible with PHP 5.6-7.2. Please update to the newer version 2.5.

PR #304, opened 2023-10-30
Strings improvement: trim ZWSP char

CrazyMuffin:
- new feature
- BC break? yes - if someone relies on having ZWSP non-trimmed while using default parameter.
- doc PR: Not needed, function is covered and does not mention the default characters

Sneaky [Zero-Width Space char \u{200B}](https://www.fileformat.info/info/unicode/char/200b/index.htm) is currently untrimmed by `trim` function, while I'd expect it to do so.

Workaround:
1. Fork or extend the class and overwrite default param
2. Manually set updated charlist as second argument to every `trim()` call

### DIFF
--- a/src/Utils/Strings.php
+++ b/src/Utils/Strings.php
@@ -21,7 +21,7 @@ class Strings
 {
 	use Nette\StaticClass;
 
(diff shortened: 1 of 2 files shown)


dg:
Thanks

PR #305, opened 2023-11-20
Added Arrays::find() method

into-the-v0id:
- new feature
- BC break? no
- doc PR: https://github.com/nette/docs/pull/1035

This PR adds the Arrays::find() helper. It's somewhat commonly used functionality and often inconvenient and unnecessarily verbose in raw PHP.

### DIFF
--- a/src/Utils/Arrays.php
+++ b/src/Utils/Arrays.php
@@ -460,4 +460,26 @@ class Arrays
 
 		return $res;
 	}
+
+
+	/**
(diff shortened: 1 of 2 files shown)


dg:
Thanks, that's a useful feature. 

I'm thinking about one thing. Probably the `findLast()` function would come in handy over time. What about utilizing the already existing `first()` and `last()` functions and just add an optional `$callback` parameter to them?

PR #311, opened 2024-05-03
Fix: 310: NormalizePath is not compatible with mikey179/vfsstream

zeleznypa:
- bug fix
- BC break no


dg:
Closing, because scheme is not part of path.

PR #314, opened 2024-06-13
Ensure valid type is passed into insert()

### LINKED ISSUE #313: Incorrect argument type definition in `addHtml` and `addText` methods
Version: latest

### Bug Description

Method `addHtml()`  has `mixed` type definition, which is incorrect. In fact it really expects `HtmlStringable|string`, because of inner call  of `->insert()`.

The same applies to `addText()`.

### Steps To Reproduce
Html::el('span')->addHtml(0); // results in following TypeError

```
TypeError Nette\Utils\Html::insert(): Argument #2 ($child) must be of type Nette\HtmlStringable|string, int given
```

### Expected Behavior
No type error being thrown

mskocik:
- bug fix #313 
- BC break? no
- doc PR: -

Making sure that `HTMLStringable` or `string` is passed into `Html::insert()` method

### DIFF
--- a/.github/workflows/coding-style.yml
+++ b/.github/workflows/coding-style.yml
@@ -7,7 +7,7 @@ jobs:
         name: Nette Code Checker
         runs-on: ubuntu-latest
         steps:
-            - uses: actions/checkout@v3
+            - uses: actions/checkout@v4
             - uses: shivammathur/setup-php@v2
               with:
                   php-version: 8.0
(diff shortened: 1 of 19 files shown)


dg:
Superseded: master resolved #313 the opposite way — `addHtml()` now has the strict signature `HtmlStringable|string` instead of casting to string internally. This PR is also based on an old branch and mixes in many unrelated changes (Cast, MemoizingIterator, Iterables::memoize/toIterator, Image refactor, readme, CI bumps). Closing.

PR #326, opened 2025-06-16
Strings::trim() trims Line Separator and Ideographic Space

vrana:
These characters occur in our data.

There are more characters listed at https://en.wikipedia.org/wiki/Whitespace_character#Unicode, maybe all of them should be included?

- bug fix
- BC break? yes
- doc PR: the exact characters are undocumented


dg:
Thanks

PR #333, opened 2025-12-05
Allow using null in addText, addHtml and insert; Closes #332

### LINKED ISSUE #332: Versioning
Hello,
this is not a bug report nor feature request.

I however would like to change the versioning if it would be possible. It is generally accepted that there are no bc breaks in minor releases. However this is second nette package that got release in past few weeks that has bc break in the minor release.

Could ve move all releases with bc breaks into major releases? So instead of v4.1 you would release v5.0 ?

What do you think about this?

juniwalk:
This is reintroducing back option to use null with Html::addText and Html::addHtml after switching from mixed type to narrower one.

- #332
- no (fixed bc break)

### DIFF
--- a/src/Utils/Html.php
+++ b/src/Utils/Html.php
@@ -573,7 +573,7 @@ class Html implements \ArrayAccess, \Countable, \IteratorAggregate, HtmlStringab
 	/**
 	 * Adds new element's child.
 	 */
-	final public function addHtml(HtmlStringable|string $child): static
+	final public function addHtml(HtmlStringable|string|null $child): static
 	{
 		return $this->insert(null, $child);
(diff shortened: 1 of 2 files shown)


dg:
I definitely don't want to add support for null.

dg:
For none of them.

juniwalk:
Alright, no point in adding that to addText, i will deal with this in my code.

juniwalk:
No no, i understand and I agree with you, i have bigger issue with the fact that it is in minor version and not major one.

//EDIT: This was in my eyes just a stopgap to undo the bc break for me and remove this in v5.0 as the null really does not make sense

PR #336, opened 2026-02-13
phpdoc for Arrays::invoke() callables params

spaze:
- bug fix / new feature?   bug fix
- BC break? no

Adds phpdoc params to `$callbacks` for `Arrays::invoke()`. Without the params PHPStan fails with errors like
>   Parameter #​1 $callbacks of static method Nette\Utils\Arrays::invoke() expects iterable<callable(): mixed>, array<callable(MichalSpacekCz\Form\UiForm, mixed): void> given.

(I wanted to use the 4.1.2 tag as a base ref for this PR, but there's no branch and GitHub doesn't let me create a PR when I select the tag as a base ref, so the base ref is the master branch but only for the PR, not for my branch. The change which added the line I'm changing is not in master. I'm not a git expert, not exactly sure what's going on here, sorry)

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -1,4 +1,4 @@
-name: Static Analysis (only informative)
+name: Static Analysis
 
 on: [push, pull_request]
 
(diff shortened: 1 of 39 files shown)


dg:
thanks, fixed

PR #337, opened 2026-04-27
Use native `clamp` function inside util

jnoordsij:
- new feature   
- no (potentially?)
- doc PR: n/a  

This replaces the manual `clamp` implementation with the PHP native `clamp`, which is new in PHP 8.6 (unreleased). See also:
- https://wiki.php.net/rfc/clamp_v2
- https://github.com/php/php-src/pull/19434
- https://github.com/symfony/polyfill/pull/554

Note that the only behavioral change is the type of exception and corresponding message. Migrating towards the new PHP native one seems like a reasonable choice to me, but if preferable I'm fine with catching it and manually throwing the old one.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,8 @@
 		}
 	],
 	"require": {
-		"php": "8.2 - 8.5"
+		"php": "8.2 - 8.5",
+		"symfony/polyfill-php86": "^1.36"
 	},
 	"require-dev": {
 		"nette/tester": "^2.5",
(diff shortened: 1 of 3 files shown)


dg:
There must be a version check using PHP_VERSION_ID.

jnoordsij:
That's why I've also added the (Symfony) polyfill; it is setup to provide the function if not natively available. I must admit the PR description was slightly short and did not clarify this entirely, but as such it should work fine both on current and future PHP versions.

dg:
🤦‍♂️

